PR #3, opened 2016-10-07
[#131355143] Add terraform code to create release buckets: datadog-agent-boshrelease

keymon:
[#131355143 Add metron metrics to datadog](https://www.pivotaltracker.com/story/show/131355143)
## What?

To keep the blobs of our custom releases, we need to create S3 buckets.

We decided to keep these buckets on the AWS CI account, as it is the one
that will run the pipelines to compile and publish the releases.

Meanwhile we don't have a better way to manage the configuation, we keep in
this repository the related repo that we can run locally.

The terraform state will be kept as commits in this repository.

In this case we create the bucket for datadog-agent-boshrelease, which we needed to fork to include the spruce blob as in https://github.com/onemedical/datadog-agent-boshrelease/pull/7
## How to review?

Check it makes sense. It has been already applied to the AWS account.
## Who?

Anyone but me or @richardc

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 vagrant/.vagrant/
 bin/fly*
+*.tfstate.backup
--- a/terraform/boshreleases/README.md
+++ b/terraform/boshreleases/README.md
@@ -1,0 +1,7 @@
+Bucket definitions for bosh releases
+====================================
+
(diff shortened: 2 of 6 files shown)


keymon:
This PR is a copy from a previous PR which we want to delete.

PR #8, opened 2016-11-09
[#131876273] Add datadog agent to Bosh and Concourse

henrytk:
## What

[#131876273 - Deploy build concourse](https://www.pivotaltracker.com/n/projects/1275640/stories/131876273)

We want to gather metrics from Bosh and Concourse so we can monitor them in Datadog. This pull request adds the Datadog agent to the machines and the necessary logic to turn off this functionality in our dev environments.

## How to review

* Set your `$DEPLOY_ENV` to a value different to your existing dev environment
* Deploy from this branch using the README. There should be no data sent to Datadog at this stage. Example:
    ```
    BOSH_INSTANCE_PROFILE=bosh-director-build CONCOURSE_INSTANCE_PROFILE=concourse-build BRANCH=add-datadog-131876273 DEPLOY_ENV="$DEPLOY_ENV" make dev bootstrap
    ```
* Follow the documentation in `doc/datadog.md` to upload credentials and deploy with checks enabled for your dev environment
* Check metrics are sent to Datadog. They should have a `bosh-deployment` tag which is set to your `$DEPLOY_ENV` value.
    * Those sent by the agent: `system.*`
    * Those sent by Riemann: `concourse.*`
* Tear down your environment afterwards (Run the destroy pipeline)

## Who can review

Anyone but me or @alext

### DIFF
--- a/doc/datadog.md
+++ b/doc/datadog.md
@@ -1,0 +1,18 @@
+# Datadog
+
+Datadog is enabled by default for non-development environments. When testing Datadog in a dev environment you must set `ENABLE_DATADOG=true` when configuring the pipelines. The pipelines are configured using `make dev bootstrap`.
+
+## Credentials
+
+When setting up an environment the Datadog credentials need to be decrypted and pushed into the S3 state bucket as they are required during pipeline runs. Each environment (dev, ci, staging, and prod) has its own set of Application and API credentials in the PaaS credential store.
+
+### Requirements
+
+* Make sure you have access to the PaaS credential store, this is required for Datadog credentials.
+* Load the AWS credentials for the environment you are setting up datadog for. These are required as the Datadog secrets file is stored in an S3 bucket.
+
+### Usage
+```
+make <ENV> upload-datadog-secrets
+```
+
--- a/Makefile
+++ b/Makefile
@@ -53,11 +53,13 @@ globals:
 dev: globals check-env-vars ## Set Environment to DEV
 	$(eval export SYSTEM_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipeline.digital)
 	$(eval export AWS_ACCOUNT=dev)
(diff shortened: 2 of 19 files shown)


henrytk:
As discussed with @alext and @saliceti offline, I have removed the commit which changed the Concourse Bosh deployment name to match the `$DEPLOY_ENV`. We are relying too much on using the `bosh-deployment` tag to identify something as being from a particular environment. This means we are going to have an issue when we have multiple deployments in the same environment, but that will have to be dealt with outside the scope of this pull request.

alext:
I've added commits to rename the bosh deployment to 'bosh', and also to add a `deploy_env` tag to both hosts.

doc/datadog.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+### Example for testing in a dev environment with standard dev API/application keys
+Upload the dev keys to your state bucket
+
+```
+make dev upload-datadog-secrets
+```
+
+### Example for testing in a dev environment with new API/application keys
```
saliceti: Is this section required? This looks like a very rare situation and I'm not sure it's worth complicating this doc.

henrytk: I have removed the section

henrytk:
A new commit has been added which adds the Concourse manifest as a resource and uses it to avoid hard coding the Concourse manifest name in the destroy pipeline. You will have to run the create pipeline at least once for existing environments for it to be uploaded to the S3 bucket.

saliceti:
The riemann metrics miss the deploy_env tag (and possibly others). Could we add it?
It can be added to the ATC job. Refer to the [spec file](https://github.com/concourse/concourse/blob/master/jobs/atc/spec#L292)

alext:
> The riemann metrics miss the deploy_env tag (and possibly others). Could we add it?

Good spot. Added and rebased.

PR #9, opened 2016-11-22
Rename the state bucket to gds-paas-${deploy_env}-state

alext:
## What

This bucket is now going to be used by the deployed concourse as well as
the bootstrap concourse, so it should be named to reflect this purpose.

This rename will be disruptive for existing deployments, but as we
haven't yet got anything running on these, it's a good time to do this.
Later on, this will be harder to change.

I've also included a small refactoring to remove an unnecessary `env` variable.

### Post-merge

Once this is merged, I will handle redeploying CI with this change.

## Suggestions for review

You could deploy a build concourse using this branch. Verify that it works, and that it uses the newly named state bucket.

## Who can review

Anyone but myself.

### DIFF
--- a/concourse/scripts/pipelines.sh
+++ b/concourse/scripts/pipelines.sh
@@ -30,7 +30,7 @@ generate_vars_file() {
 aws_account: ${AWS_ACCOUNT}
 vagrant_ip: ${VAGRANT_IP}
 deploy_env: ${env}
-state_bucket: gds-paas-${env}-bootstrap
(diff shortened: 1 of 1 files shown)


alext:
Note: This depends on https://github.gds/government-paas/aws-account-wide-terraform/pull/66 being merged before it can be deployed to CI.

benhyland:
lgtm, Alex has a convincing argument that he knows what he is doing ;)

PR #10, opened 2016-11-23
Add make target to setup build profiles.

alext:
## What

To save looking up the relevant instance profiles each time, this adds a
make target to set them. This can be used as `make dev build-concourse
bootstrap`. I anticipate that other profiles will be added over time as
this expands to handling more than just the build concourse.

## How to review

Check the changes look sensible.
Check you can spin up a bootstrap concourse using the new target (no need to run any pipelines in this).

## Who can review

Anyone but myself.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,8 @@ globals:
 	$(eval export AWS_DEFAULT_REGION=eu-west-1)
 	@true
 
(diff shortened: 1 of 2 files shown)


benhyland:
not just a pretty face

PR #12, opened 2016-12-09
[#133036569] Fix specification of ELB security policy.

alext:
## What

This fixes #11. The previous version will only work for recently created ELBs. See the commit message for more details.

## How to review

This can be reviewed in the same manner as https://github.com/alphagov/paas-cf/pull/653.

Alternatively, you could trust that if the change works on paas-cf, it'll work here.

## Who can review

Anyone but myself.

### DIFF
--- a/terraform/concourse/elb.tf
+++ b/terraform/concourse/elb.tf
@@ -34,13 +34,15 @@ resource "aws_elb" "concourse" {
   }
 }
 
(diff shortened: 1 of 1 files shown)


keymon:
I only did a code review but I did not run it. The code is exactly the same as  https://github.com/alphagov/paas-cf/pull/653, and makes sense. 

merging

PR #15, opened 2016-12-16
[#131876783] Merge shellcheck and terraform fmt changes from paas-cf

dcarley:
## What

Update with some recent changes from alphagov/paas-cf. This will make it easier to sync Terraform changes before the code is removed from paas-cf, which are coming up next.

## How to review

1. Confirm that the Travis build passes.
1. Confirm that there don't appear to be any functional changes to the Terraform code. Viewing with `git show -b` or adding `?w=0` to the PR will make this easier.
1. I don't think it's necessary to run the whole paas-bootstrap pipeline for these changes, but happy to take a second opinion.

## Who can review

Not @dcarley. It's probably OK for @alext to review, since I did this independently.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ rvm:
   - 2.2.3
 env:
   global:
-    - TF_VERSION="0.7.3"
+    - TF_VERSION="0.7.10"
     - SPRUCE_VERSION="1.5.0"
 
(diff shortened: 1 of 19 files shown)


mtekel:
LG. As it turns out, ?w=1 does the same thing as ?w=0...

PR #16, opened 2016-12-16
[#126126973] Allow STS creds to be used to bootstrap

benhyland:
## What

Story: [Ensure all AWS IAM accounts have MFA enabled](https://www.pivotaltracker.com/story/show/126126973)

We need to allow short-lived credentials to be used in order
to require MFA for aws users.

See https://www.pivotaltracker.com/story/show/126126973 https://www.pivotaltracker.com/story/show/136052945 and  https://www.pivotaltracker.com/story/show/136052607

The sts-generated session token needs to be passed to vagrant.

The concourse atc password can no longer be generated from
aws creds, since they are no longer static but may change each day
when short-lived credentials are regenerated.
For the bootstrap pipeline, we can use a static atc password, since the
only access is via ssh - the concourse is not exposed to the
public internet. But instead of using the AWS key as before, we now use
the AWS UserId.

We explicitly separate the two concourse passwords to make this
easier to understand and maintain in future.

## How to review

Deploy a new env using the bootstrap pipeline.
Check that the concourse lite and deployer concourse passwords work as you expect.

## Who can review

Anyone, but it might be nice to get a quick once-over from @alext since he's doing other changes in this area.

Note: This should only be merged after the work to migrate permanent environments to use the `pass-bootstrap` pipeline has been deployed.


alext (changes requested):
There's a paragraph in the README (here https://github.com/alphagov/paas-bootstrap/blob/1514c6f52b372ab59db2485c21c1b6dc05d97cc5/README.md#deploy-1) that still says that the deployer concourse password is the same as the cocourse-lite password.

It would be good to clarify what the migration process for the long-lived environments will be for this change. Will they continue to use the existing passwords (as stored in paas-pass), or will they use newly generated ones?

concourse/scripts/deployer-environment.sh
```diff
@@ -0,0 +1,31 @@
...
+fi
+
+AWS_ACCOUNT=${AWS_ACCOUNT:-dev}
+
+CONCOURSE_ATC_USER=${CONCOURSE_ATC_USER:-admin}
+CONCOURSE_URL="${CONCOURSE_URL:-https://concourse.${SYSTEM_DNS_ZONE_NAME}}"
+FLY_TARGET=${FLY_TARGET:-$DEPLOY_ENV}
+FLY_CMD="${PROJECT_DIR}/bin/fly"
+
+CONCOURSE_ATC_PASSWORD=$(aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/concourse-secrets.yml" - | grep "concourse_atc_password" | cut -d ':' -f 2 | tr -d ' ')
```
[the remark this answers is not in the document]
alext: Having discussed this I think this is fine to merge as is because this pattern is currently used in several other places. Recent changes to how we generate passwords also mean that they should always be unquoted.

benhyland:
@alext
I fixed the readme.
As discussed, we think the hacky yml 'parsing' in bash will be OK for now since we have Vault/a.n.other secret passing mechanism in the backlog.
We need to applied whatever migration is needed to move permanent envs onto the paas-bootstrap pattern before applying this change.

benhyland:
rebased against master following work related to migrating paas-cf to paas-bootstrap

combor:
This is ready for review again.

concourse/scripts/concourse-lite-environment.sh
```diff
@@ -0,0 +1,37 @@
...
+
+AWS_ACCOUNT=${AWS_ACCOUNT:-dev}
+
+CONCOURSE_URL="${CONCOURSE_URL:-http://localhost:8080}"
+FLY_TARGET="${FLY_TARGET:-${DEPLOY_ENV}-bootstrap}"
+FLY_CMD="${PROJECT_DIR}/bin/fly-bootstrap"
+
+CONCOURSE_ATC_USER=${CONCOURSE_ATC_USER:-admin}
+if [ -z "${CONCOURSE_LITE_ATC_PASSWORD:-}" ]; then
+    CONCOURSE_LITE_ATC_PASSWORD=$(hashed_password "$(gpg --export-secret-key -a)")
```
dcarley: It feels uncomfortable using my private key in this way. I can't describe the exact problem, but it *feels* wrong. We had similar concerns about `AWS_ACCESS_KEY` in the very first implementation.

The original PR used a less secure predictable string on the basis that the bootstrap Concourse can only be accessed using an SSH tunnel. Could you describe the rationale for changing it?

dcarley: Some of us talked about this at the desks. There's discomfort about the Security Group being the only restriction to accessing the bootstrap Concourse. We could also change the bind address for ATC so that it's only accessible over loopback (SSH tunnel).

However @Jonty suggested that we could use something from the AWS API that's not easily determinable by people outside the team - such as account ID, user ARN, or user ID. @combor and @saliceti are looking into it.

alext (approved):

concourse/scripts/concourse-lite-environment.sh
dcarley: Resolved.

concourse/scripts/concourse-lite-environment.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
keymon: If we add a new file `concourse-lite-environment.sh`, why do we need `environment.sh` at all in paas-bootstrap.

Anyway, I am not 100% confortable that both scripts are different

vagrant/Vagrantfile
```diff
@@ -41,6 +41,7 @@ Vagrant.configure(2) do |config|
   config.vm.provider :aws do |aws, override|
     aws.access_key_id = ENV['AWS_ACCESS_KEY_ID']
     aws.secret_access_key = ENV['AWS_SECRET_ACCESS_KEY']
+    aws.session_token = ENV['AWS_SESSION_TOKEN']
```
keymon, line 44: I would have split this from the main commit, as it is a really specific change not related with the rest. But it does not matter ;)

concourse/scripts/environment.sh
```diff
@@ -9,38 +9,22 @@ hashed_password() {
...
-  if [ -n "${DECRYPT_CONCOURSE_ATC_PASSWORD:-}" ]; then
-    CONCOURSE_ATC_PASSWORD=$(pass "${DECRYPT_CONCOURSE_ATC_PASSWORD}/concourse_password")
-  else
-    CONCOURSE_ATC_PASSWORD=$(hashed_password "${AWS_SECRET_ACCESS_KEY}:${DEPLOY_ENV}:atc")
-  fi
+    user_id=$(aws sts get-caller-identity | awk '$1 ~ /UserId/ {print $2}')
+    CONCOURSE_ATC_PASSWORD=$(hashed_password "${user_id}")
 fi
 
 cat <<EOF
```
keymon, line 30: You could add here a warning above `export CONCOURSE_ATC_PASSWORD=${CONCOURSE_ATC_PASSWORD}`:

```
# WARNING: This is the password for the Bootstrap Concourse Lite in paas-bootstrap.
export CONCOURSE_ATC_PASSWORD=qwertyuiop234567890
```

vagrant/deploy.sh
```diff
@@ -7,9 +7,8 @@ PROJECT_DIR=$(cd "${SCRIPT_DIR}/.." && pwd)
 cd "$SCRIPT_DIR"
 
 # Load environment variables
-export TARGET_CONCOURSE=bootstrap
 # shellcheck disable=SC2091
-$("${PROJECT_DIR}/concourse/scripts/environment.sh")
+$("${PROJECT_DIR}/concourse/scripts/environment.sh" "$@")
```
keymon, line 11: You must add here `eval` before, or otherwise you get the error:

```
$ BRANCH=$(git rev-parse --abbrev-ref HEAD) DEPLOY_ENV=hector make dev deployer-concourse bootstrap 
vagrant/deploy.sh
vagrant/deploy.sh: line 11: export: `#': not a valid identifier
vagrant/deploy.sh: line 11: export: `WARNING:': not a valid identifier
vagrant/deploy.sh: line 11: export: `paas-bootstrap': not a valid identifier
make: *** [bootstrap] Error 1
```

You can check it by running:

```
$(DEPLOY_ENV=hector /home/keymon/gds/paas-bootstrap/concourse/scripts/environment.sh)
``` 
vs 

``` 
eval $(DEPLOY_ENV=hector /home/keymon/gds/paas-bootstrap/concourse/scripts/environment.sh)
```

keymon: Actually it does not work because the comment with the warning. It seems that all the part after the comment is ignored.

we could do this:
```
diff --git a/concourse/scripts/environment.sh b/concourse/scripts/environment.sh
index fc2261b..153d5e0 100755
--- a/concourse/scripts/environment.sh
+++ b/concourse/scripts/environment.sh
@@ -28,10 +28,12 @@ if [ -z "${CONCOURSE_ATC_PASSWORD:-}" ]; then
 fi
 
 cat <<EOF
-export AWS_ACCOUNT=${AWS_ACCOUNT}
+export AWS_ACCOUNT=${AWS_ACCOUNT};
 export DEPLOY_ENV=${DEPLOY_ENV}
 export CONCOURSE_ATC_USER=${CONCOURSE_ATC_USER}
-# WARNING: This is the password for the Bootstrap Concourse Lite in paas-bootstrap
+EOF
+echo '# WARNING: This is the password for the Bootstrap Concourse Lite in paas-bootstrap' 1>&2
+cat <<EOF
 export CONCOURSE_ATC_PASSWORD=${CONCOURSE_ATC_PASSWORD}
 export CONCOURSE_URL=${CONCOURSE_URL}
 export FLY_CMD=${FLY_CMD}
``` 

Or otherwise, simply remove the comment.

combor: Warning removed.

keymon:
LGTM now. Merging

concourse/scripts/environment.sh
```diff
@@ -9,38 +9,22 @@ hashed_password() {
   echo "$1" | shasum -a 256 | base64 | head -c 32
 }
 
-DEPLOY_ENV=${DEPLOY_ENV:-}
+DEPLOY_ENV=${1:-${DEPLOY_ENV:-}}
```
alext, line 12: Why is this positional parameter being re-introduced? I removed this in 5a70bcd because it's not used anywhere...

saliceti: @alext good catch
Here you go: https://github.com/alphagov/paas-bootstrap/pull/35

PR #19, opened 2016-12-19
Fix state bucket name in dev scripts.

alext:
## What

These references to the old state bucket name were missed when the
bucket was renamed in ead368e. Without this, ssh to the concourse
instance fails because it can't find the key in the bucket etc.

## How to review

Attempt to ssh to a concourse deployed from this repo. It should now succeed

## Who can review

Anyone but myself.


benhyland:
Failed before, succeeds after. yay :)

PR #20, opened 2016-12-19
[#131876783] Parameterise CONCOURSE_HOSTNAME

dcarley:
## What

Add a new CONCOURSE_HOSTNAME variable so that we can deploy Concourse with a
different hostname. The default for the `build-concourse` target remains the
same.

This change is necessary because we're planning to remove the bootstrap code
from alphagov/paas-cf and use alphagov/paas-bootstrap instead, however
paas-cf expects a Concourse with the hostname of `deployer.${domain}`
instead of `concourse.${domain}`.

We can change the URL in the README when we come to introduce a new target.

## How to review

1. Make a bootstrap Concourse with: `make dev build-concourse bootstrap`
1. Confirm that you can login and it has the right TLS cert at: `https://concourse.${DEPLOY_ENV}.dev.cloudpipeline.digital/`

## Who can review

Not @dcarley or @alext


benhyland:
lgtm

PR #21, opened 2016-12-19
[#131876783] Make non-dev Concourses safer to use

dcarley:
## What

Port some changes from alphagov/paas-cf which make Concourse servers in non-development environments safer to use, by introducing a "team" which indicates the environment name and forcing people to re-login after short periods of time when making changes.

## How to review

This functionality has been tested before in paas-cf and I've tested it from scratch against this branch (in addition to ongoing WIP branch that Alex and I have for Deployer Concourse), so if you're comfortable with reviewing the code-only then I think that'd be OK.

1. Checkout this branch: `feature/131876783-distinct_concourses`
1. Create a bootstrap: `BRANCH=$(git rev-parse --abbrev-ref HEAD) make dev build-concourse bootstrap`
1. Run the `create` pipeline.
1. Confirm that it shows you the environment name when you login at: `https://concourse.${DEPLOY_ENV}.dev.cloudpipeline.digital/`

Testing the auth duration is more difficult. You can temporarily remove the line `$(eval export CONCOURSE_AUTH_DURATION=48h)` from the `dev` target in the `Makefile` and repeat the steps, then see if you have to re-login to the Build Concourse after 5 minutes.

## Who can review

Not @dcarley

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,7 @@ dev: globals check-env-vars ## Set Environment to DEV
 	$(eval export SYSTEM_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipeline.digital)
 	$(eval export AWS_ACCOUNT=dev)
 	$(eval export ENABLE_DATADOG ?= false)
+	$(eval export CONCOURSE_AUTH_DURATION=48h)
 
 .PHONY: ci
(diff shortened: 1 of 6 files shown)


benhyland:
looks grand.

PR #22, opened 2016-12-20
[#131876783] Remove vagrant IP from bosh_rds security group.

alext:
## What

There's no need for concourse-lite to connect to the RDS instance.
Additionally, this entry wasn't being expunged at the end of the
pipeline because the expunge terraform run only targets the `bosh`
security group, and doesn't include this one.

## How to review

Run the create pipeline and verify it successfully deploys a bosh and concourse.

## Who can review

Anyone but myself.

### DIFF
--- a/terraform/bosh/rds.tf
+++ b/terraform/bosh/rds.tf
@@ -14,10 +14,9 @@ resource "aws_security_group" "bosh_rds" {
   vpc_id      = "${var.vpc_id}"
 
   ingress {
-    from_port   = 5432
-    to_port     = 5432
-    protocol    = "tcp"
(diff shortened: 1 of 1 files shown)


saliceti:
I thought terraform needed access to port 5432?
Terraform runs in concourse-lite to create that RDS instance.

alext:
> I thought terraform needed access to port 5432?

I don't think it does. I think it purely operates using the AWS API. There may have been a time when that was necessary when we were attempting to use the postgresql provider to initialise the database. I think that was only for the CF database though, not the Bosh one.

saliceti:
You're right, it was for the CF DB.
Merging

PR #24, opened 2016-12-20
[#131876783] Enable datadog bosh health_manager integration

alext:
## What

This pull in the change that was made in paas-cf here: https://github.com/alphagov/paas-cf/pull/607

Bosh health_manager supports integration with datadog[1].
It sends all metrics received during the agent heartbeats.

These heartbeat happen once every minute, and include multiple metrics
about the VM health: cpu, load, memory... and about the monit services in
the VM being running OK (`bosh.healthmonitor.system.healthy`)

We are mainly interested in `bosh.healthmonitor.system.healthy` as the
other metrics are also reported by the datadog agents, which we will not
stop using.

We use the variable ENABLE_DATADOG to enable/disable the datadog
integration.

[1] https://github.com/cloudfoundry/bosh/blob/master/release/jobs/health_monitor/spec#L185-L194

## How to review

Code review - verify that these changes match the ones in paas-cf that they're based on.

As per https://github.com/alphagov/paas-cf/pull/607, deploy with Datadog enabled and verify that the checks are created correctly.

## Who can review

Anyone but myself or @dcarley

### DIFF
--- a/concourse/pipelines/create.yml
+++ b/concourse/pipelines/create.yml
@@ -585,6 +585,7 @@ jobs:
           params:
             AWS_ACCOUNT: {{aws_account}}
             DATADOG_API_KEY: {{datadog_api_key}}
+            DATADOG_APP_KEY: {{datadog_app_key}}
             ENABLE_DATADOG: {{enable_datadog}}
             BOSH_MANIFEST_STUBS: |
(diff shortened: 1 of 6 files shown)


saliceti:
LGTM

PR #26, opened 2016-12-22
[#131876783] Migration from paas-cf to paas-bootstrap

keymon:
[#131876783 Migrate paas-cf to paas-bootstrap](https://www.pivotaltracker.com/story/show/131876783)

**NB: Please review but do not merge over holiday period**

What?
-----

We want to use paas-bootstrap project to create the initial VPC, bosh and concourse deployer, and remove the required logic from paas-cf.

For that, apart of migrate the logic, we must also implement a procedure to migrate the existing environments.

In this PR we :
 * Move the basic configuration to get VPC, bosh and concourse running from paas-cf.
 * Implement logic to migrate an existing environment:
   * Copy blobstores across, as they get renamed.
   * Recreate the existing bosh VMs to allow change the blobstore.
   * Reattach the old volume from concourse into the new one.

All the migration steps are idempotent and the commits can be reverted later after the migration is done in all the environments.

Note: There are several features and small useful scripts that were not migrated yet. The change is big enough and will be done in follow up PRs.

Dependencies
-----------------

This PR must be reviewed, tested and applied in conjuction with https://github.com/alphagov/paas-cf/pull/688

How to review, test and apply it?
---------

Do a code review to check that makes sense.

To test it: 

 1. Deploy a CF environment using the master branch of  https://github.com/alphagov/paas-cf
 2. Apply the procedure described below.

Migration procedure
------------------------

This branch includes all the required logic almost do all the migration from a paas-cf deployment. The steps are in the commits prefixed with `migrate`.

The full migration process would be:

1. Optionally pause all pipelines in the Deployer Concourse. eg:
   ```
   TARGET=... # master, prod, staging...
   ./bin/fly -t ${TARGET} pipelines -a | grep main | awk '{print $1}' | xargs -n1 -I {} ./bin/fly -t ${TARGET} pause-pipeline -p {}
   ``` 
2. Bring up new Bootstrap Concourse from `alphagov/paas-bootstrap`
    ```
    DEPLOY_ENV=... BRANCH=$(git rev-parse --abbrev-ref HEAD) make dev deployer-concourse bootstrap
    ```
1. Run the `create` pipeline on the Bootstrap Concourse.
   This will automatically: 
   	 * Create the new buckets and sync them.
   	 * Update the terraform resources (eg. vpc name)
   	 * update BOSH (new blobstore bucket)
   	 * migrate CF VMs for new blobstore, 
   	 * delete old concourse
   	 * recreate Concourse (takes approx 1hr)
   	 * attach the old disk to concourse and update bosh DB.
    
   Note: This failed for me once when updating `parser_z2/0` because it was unable to connect to Elasticsearch. The simplest, but time consuming, solution is to re-run the job.

1. Delete the cookies from your deployer. We found out that you cannot login after re-deployment having the old cookie set (???).

1. Run the  migration from paas-cf PR https://github.com/alphagov/paas-cf/pull/688:  Run `BRANCH=$(git rev-parse --abbrev-ref HEAD) make dev pipelines DEPLOY_ENV=....` (in https://github.com/alphagov/paas-cf/pull/688) to update all the pipelines in the new concourse (this will unpause them as part of the deployment).

Once this is finished:

1. Clean up manually:
	* Delete old S3 state bucket `<env>-state`. It's versioned, so you'll need to use the AWS console.
	* Delete old S3 blobstore bucket. `aws s3api delete-bucket --bucket ${DEPLOY_ENV}-bosh-blobstore`
1. Once all environments have been migrated (all), we can revert all the commits prefixed with `migrate:` in paas-bootstrap and paas-cf
1. ???
1. Profit

Who?
----

Anyone but @keymon, @dcarley or @alext

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,13 @@ build-concourse: ## Setup profiles for deploying a build concourse
 	$(eval export CONCOURSE_INSTANCE_PROFILE=concourse-build)
 	@true
 
+.PHONY: deployer-concourse
+deployer-concourse: ## Setup profiles for deploying a paas-cf deployer concourse
+	$(eval export BOSH_INSTANCE_PROFILE=bosh-director-cf)
+	$(eval export CONCOURSE_HOSTNAME=deployer)
+	$(eval export CONCOURSE_INSTANCE_PROFILE=deployer-concourse)
+	@true
+
 ## Actions
 
 .PHONY: fly-login
--- a/Makefile
+++ b/Makefile
@@ -50,8 +50,10 @@ lint_ruby:
 	bundle exec govuk-lint-ruby
 
 .PHONY: globals
+PASSWORD_STORE_DIR?=${HOME}/.paas-pass
 globals:
 	$(eval export AWS_DEFAULT_REGION=eu-west-1)
+	$(eval export PASSWORD_STORE_DIR=${PASSWORD_STORE_DIR})
 	@true
 
 ## Environments
@@ -68,18 +70,21 @@ ci: globals check-env-vars ## Set Environment to CI
 	$(eval export SYSTEM_DNS_ZONE_NAME=${DEPLOY_ENV}.ci.cloudpipeline.digital)
 	$(eval export AWS_ACCOUNT=ci)
 	$(eval export ENABLE_DATADOG=true)
+	$(eval export DECRYPT_CONCOURSE_ATC_PASSWORD=ci_deployments/${DEPLOY_ENV})
 
 .PHONY: staging
 staging: globals check-env-vars ## Set Environment to Staging
 	$(eval export SYSTEM_DNS_ZONE_NAME=staging.cloudpipeline.digital)
 	$(eval export AWS_ACCOUNT=staging)
 	$(eval export ENABLE_DATADOG=true)
+	$(eval export DECRYPT_CONCOURSE_ATC_PASSWORD=staging_deployment)
 
 .PHONY: prod
 prod: globals check-env-vars ## Set Environment to Production
 	$(eval export SYSTEM_DNS_ZONE_NAME=cloud.service.gov.uk)
 	$(eval export AWS_ACCOUNT=prod)
 	$(eval export ENABLE_DATADOG=true)
+	$(eval export DECRYPT_CONCOURSE_ATC_PASSWORD=prod_deployment)
 
 
 ## Concourse profiles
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -68,6 +70,21 @@ ci: globals check-env-vars ## Set Environment to CI
...
+.PHONY: staging
+staging: globals check-env-vars ## Set Environment to Staging
+	$(eval export SYSTEM_DNS_ZONE_NAME=staging.cloudpipeline.digital)
+	$(eval export AWS_ACCOUNT=staging)
+	$(eval export ENABLE_DATADOG=true)
+	$(eval export DECRYPT_CONCOURSE_ATC_PASSWORD=staging_deployment)
+
+.PHONY: prod
+prod: globals check-env-vars ## Set Environment to Production
+	$(eval export SYSTEM_DNS_ZONE_NAME=prod.cloudpipeline.digital)
```
saliceti: Isn't it `cloud.service.gov.uk` for prod?

concourse/scripts/migrate_concourse_volume.sh
```diff
@@ -0,0 +1,377 @@
...
+      echo "Attaching ${origin_concourse_volume_id} to ${new_concourse_instance_id} as /dev/sdf"
+      attach_volume "${origin_concourse_volume_id}" "${new_concourse_instance_id}" "/dev/sdf"
+    fi
+  fi
+
+  echo "Starting again the new concourse instance ${new_concourse_instance_id}"
+  start_instance "${new_concourse_instance_id}"
+  wait_instance_state "${new_concourse_instance_id}" "running"
+
+  # TODO: Check if we need to do a monit restart all, not sure if it really starts :-m
```
saliceti, line 299: It works for me without restart all, this comment can be removed

saliceti:
Could we prefix commits https://github.com/alphagov/paas-bootstrap/pull/26/commits/1980dfa728e5d346c31f4a78f091742ec16e7016 and https://github.com/alphagov/paas-bootstrap/pull/26/commits/f59ab96d23fda4a1520b8bba4db582ec9c312701 with "migrate: "? I believe they are required for the migration only and this will be easier to identify them so we can revert them later.

saliceti:
The migration worked superbly. A couple of notes:

* Initially I had a state bucket left over from another paas-bootstrap and the pipeline failed with `BucketAlreadyOwnedByYou`. I deleted it manually and it worked fine.
* There are error messages like `CONCOURSE_HOSTNAME: unbound variable` that should be ignored in `migrate-concourse-step2`. This is because `showenv` is run to get the concourse ip, the rest is run and fails but it is not required.

Makefile
alext: It is.  Fixed and rebased.

concourse/scripts/migrate_concourse_volume.sh
alext: I'm not sure it's worth removing this comment as this script is going to be removed once the migration is complete.

PR #27, opened 2017-01-03
[#131876783] Remove datadog-agent from concourse manifest and skip migration on new deployments.

alext:
## What

This conflicts with the job defined in the runtime config created by
paas-cf. This means that the deployment will fail when using a bosh
that has had the runtime-config from paas-cf applied.

This is a temporary fix to unblock the deployment. Once the deployment is 
complete we'll work on a proper fix for this which will probably involve migrating\
the runtime config to this repo.

Additionally, we add logic to do not try to migrate if the state is missing. 

## How to review

Code review. Make sure this makes sense.

## Who can review

Anyone but @keymon or myself.


mtekel:
migration skip works, now trying to confirm datadog removal really works

mtekel:
I think we can also drop https://github.com/alphagov/paas-bootstrap/blob/master/concourse/pipelines/create.yml#L1018 - as this now only adds datadog-agent release to concourse manifest, which is not used anywhere in the manifest

we could also rename ./paas-bootstrap/manifests/concourse-manifest/extensions/datadog-agent.yml to st. like 'rieman-for-datadog' or 'concourse-metrics' - as it has no dog agent any more

alext:
@mtekel Those comments make sense, but can we merge it as is as this is only a temporary fix to unblock things.

We're going to come back and fix this properly as part of this story once the deployment is complete.

mtekel:
merging then...

PR #28, opened 2017-01-03
[#131876783] Add missing prod and staging tfvars files.

alext:
## What

These were missed when adding the Makefile targets in 14f78c8

## How to review

Code review. Verify that the values match the ones in paas-cf.

## Who can review

Anyone but myself.


mtekel:
the env. names and zone IDs match those in paas-cf, merging (once CI finishes)

PR #29, opened 2017-01-03
[#131876783] Add bosh DB params to prod/staging tfvars.

alext:
## What

These were omitted in c028528, and without them the bosh deb gets
created non-HA etc.

## How to review

Code review.

## Who can review

Anyone but myself.


mtekel:
LG, merging

PR #34, opened 2017-01-05
[#131876783] Move runtime-config from paas-cf to paas-bootstrap

keymon:
[#131876783 Migrate paas-cf to paas-bootstrap](https://www.pivotaltracker.com/story/show/131876783)

What?
-----

We split the deployment of bosh and concourse in https://github.com/alphagov/paas-cf/pull/688

But we did not move the [runtime-config](https://bosh.io/docs/runtime-config.html). This configuration is global for all the bosh deployments. This causes a  cycled dependency if we keep runtime-config in the paas-cf project: the first deployment of concourse will not include runtime-config, but paas-cf will add it later.

It makes more sense to transfer all the code to the paas-bootstrap repository and manage it from there.

We move all the configuration for collectd, syslog_forwarder and datadog.

Some notes:

 * for clarity, we restructure the files to manifests/runtime-config/addons  and manifests/runtime-config/addons-meta.

 * collectd: The configuration points to a hardcoded IP. We will maintain that for the time being.

  * syslog-forwarder: The syslog target (logstash from logsearch) is deployed in paas-cf. We will point to a generated DNS name, which will be created/updated during deployment of logsearch.

  * Datadog: the configuration was already in this project, but we change the path of the files included. We will update the bosh and concourse manifest laters. We get the datadog api from environment variables.

  * We want to be able to toggle some the addons, like collectd and syslog which are only enabled for CF deployer. For that, we keep the configuration in multiple files that will be conditionally added during the manifest generation.

Dependencies
-----------

To review with related paas-cf PR: https://github.com/alphagov/paas-cf/pull/703

How to review?
-------------

 * Code review.
 * Travis should pass.
 * Run the bootstrap pipeline as usual.
    * The `generate-bosh-config` and `deploy-bosh` should generate and upload the runtime config as desired, and all the releases.
    * concourse should deploy all the addons.
    * bosh should deploy collectd

Who?
---

Anyone but @keymon or @alext

[#131876783 Migrate paas-cf to paas-bootstrap](https://www.pivotaltracker.com/story/show/131876783)


alext:
I'm currently investigating whether we can import the history for these files from paas-cf. The final state will remain the same, so this can still be reviewed in the meantime.

alext:
I've updated this to include the history from paas-cf.

keymon:
@alext that is great! good job! :+1:

mtekel:
lg, merging

PR #39, opened 2017-01-17
[#137698283] Generate secrets in tests

alext:
## What

To avoid having to maintain secrets fixtures, this updates the tests to use the secret generation scripts to generate these as needed.

As part of this work I've also taken the opportunity to clean up some other details in these tests. See individual commits for details.

## How to review

Code review. Verify tests pass in Travis.

## Who can review

Anyone but myself.


manifests/concourse-manifest/spec/manifest_generation_spec.rb
```diff
@@ -34,18 +33,10 @@ def merge_fixtures(fixtures)
...
-
-    expect(
-      manifest_with_defaults["jobs"][0]["templates"][1]["properties"]["basic_auth_password"]
-    ).to eq(fixtures["secrets"]["concourse_atc_password"])
-  end
-
-  it "gets values from predefined secrets" do
+  it "gets values from secrets" do
     expect(
       atc_template.fetch("properties").fetch("basic_auth_password")
```
dcarley, line 38: Out of interest, are `manifest_with_defaults["jobs"][0]["templates"][1]["properties"]["basic_auth_password"]` (old test) and `atc_template.fetch("properties").fetch("basic_auth_password")` (new test) the same thing?

I couldn't tell from 26ca5f043cab05df5b7c4b8f06a49a8c9c226685 either.

alext: They are. If you look at https://github.com/alphagov/paas-bootstrap/blob/master/manifests/concourse-manifest/concourse-base.yml#L78 you can see that `jobs[0]` Is the concourse job, and `templates[1]` is the atc template on that job. Which is what the `atc_template` rspec helper pulls out at the top of this file (although it pulls them out by name, which is much more robust.)

dcarley: Cool, thanks 👍

PR #41, opened 2017-01-20
[Support] Use a terraform array for admin_cidrs

alext:
## What

Makes it clearer and avoids an extra interpolation function. This also makes it consistent with paas-cf.

## How to review

Code review (compare with https://github.com/alphagov/paas-cf/commit/965ae1e).
You could run this through the pipeline and verify that the change is a no-op (other than the Vagrant IP being added and removed as usual).

## Who can review

Anyone but myself.


saliceti:
Works fine

PR #46, opened 2017-02-09
[#13435058] Enable TLS for syslog

combor:
# What

As we would like to have be compliant with AWS Trusded advisor
recommendations we need to enable TLS communication to syslog ingestor
ELB.

## How to review

- Review https://github.com/alphagov/paas-cf/pull/776 first and deploy it on top of master. It creates new ELB listeners for TLS syslog.
- run  bootstrap from this branch and check if logs are still visible in Kibana

## Who can review

Not @paroxp or @combor


keymon:
Merging, I will apply this in ci, staging and prod

PR #48, opened 2017-02-20
[#119379405] Upgrade bootstrap concourse 

mtekel:
## What

[Upgrade bootstrap concourse ](https://www.pivotaltracker.com/story/show/119379405)
Use latest official docker container and recommended docker compose way of deploying it to get latest official concourse.

## How to review

Set up your usual variables (DEPLOY_ENV, AWS credentials). Checkout this branch, do

```
export BRANCH=$(git rev-parse --abbrev-ref HEAD)
make dev deployer-concourse bootstrap
```

Observe vagrant starting a new VM and then docker compose running on it. Tunnel is being created and pipelines uploaded. Run the pipeline - it should not apply changes to your existing environment. Meanwhile, have a look at README changes - proofread.

## Who can review

not @mtekel or @paroxp


saliceti:
The first time I ran it the tunnel failed:

```
Setting up SSH tunnel to concourse...
Failed creating SSH tunnel to remote concourse: 'vagrant ssh -- -L 8080:127.0.0.1:8080 -N'
```

Then it worked the second time. Can we avoid this issue?

mtekel:
Yes we know about it and have documented it: https://github.com/alphagov/paas-bootstrap/pull/48/files#diff-04c6e90faac2675aa89e2176d2eec7d8R229

vagrant/post-deploy.d/00-run-docker.sh
```diff
@@ -0,0 +1,27 @@
...
+ssh-keygen -t rsa -f ./keys/worker/worker_key -N ''
+
+cp ./keys/worker/worker_key.pub ./keys/web/authorized_worker_keys
+cp ./keys/web/tsa_host_key.pub ./keys/worker
+
+# shellcheck disable=SC2091
+$("${PAAS_DIR}/concourse/scripts/environment.sh")
+
+sudo -E docker-compose up &
+sleep 60 # Let it start up
```
saliceti: Could we use a smart loop instead of sleep?

mtekel: we could do that, but that sleep has to stay there for compose to correctly boot up (it seems to need the terminals vagrant provides during the run of the provisioning scripts)

vagrant/post-deploy.d/00-run-docker.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
+
+set -eu
+
+PAAS_DIR=/tmp/paas-bootstrap
+
+sudo apt update && sudo apt install docker-compose -y
```
saliceti: Have we looked into using an AMI with docker included so we can avoid any installation?

mtekel: We thought about CoreOS, but then decided to use most common and most default image so that we can avoid any possible (obscure) compatibility issues into the future and also to ensure image will be available for long time. We chose ubuntu 16.04 as its support is until 2020...

docker/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
+concourse-db:
```
saliceti, line 1: Can we pin all (postgres and concourse) container image versions?
We want reproducible deployments.

paroxp: We can. Please take a look at [my comment](https://www.pivotaltracker.com/story/show/119379405/comments/163752543) on the pivotal story and give me final shout.

saliceti: I did read the comments in pivotal. As far as I understand, we want to automate the deployment of paas-bootstrap. If concourse or postgres docker images are updated between dev and prod, we may end up with different versions in different accounts.

vagrant/post-deploy.d/00-run-docker.sh
```diff
@@ -0,0 +1,27 @@
...
+
+git clone https://github.com/alphagov/paas-bootstrap.git ${PAAS_DIR}
```
saliceti: The vagrant directory is available on the vagrant box. If you copy the required files here, no need to use git.

vagrant/post-deploy.d/00-run-docker.sh
```diff
@@ -0,0 +1,27 @@
...
+ssh-keygen -t rsa -f ./keys/web/session_signing_key -N ''
+ssh-keygen -t rsa -f ./keys/worker/worker_key -N ''
+
+cp ./keys/worker/worker_key.pub ./keys/web/authorized_worker_keys
+cp ./keys/web/tsa_host_key.pub ./keys/worker
+
+# shellcheck disable=SC2091
+$("${PAAS_DIR}/concourse/scripts/environment.sh")
+
+sudo -E docker-compose up &
```
saliceti: You can use switch `-d` so docker will be in daemon mode after it downloads the images. That way you can avoid the `&` and docker is ready when the script exits.
We still need to wait for ATC to start.

vagrant/Vagrantfile
```diff
@@ -44,13 +45,16 @@ Vagrant.configure(2) do |config|
     aws.session_token = ENV['AWS_SESSION_TOKEN']
     aws.associate_public_ip = true
     aws.tags = { 'Name' => (ENV['DEPLOY_ENV'] || ENV['USER']) + " concourse" }
+    aws.keypair_name = ENV['VAGRANT_SSH_KEY_NAME']
```
dcarley, line 48: Is the addition of a separate SSH key required? If so, could you explain why in the commit message?

paroxp: It's required, as we have deprecated a Vagrant box image we've been using thus far, which has been a resource of an unknown origin and we've been worried it will disappear anytime. Other side of the medal, is the fact that the box had "hardcoded" ssh key, which could mean, for the time the box is up anyone with that key would have an access to the VM in the time of its live.

With our new approach, we're creating the key on the go with AWS Key Pairs and then removing it in the destroy action. We're also able, to use official box (Ubuntu) as oppose to rely on third-party images.

docker/docker-compose.yml
dcarley: 👍  pinning versions for reproducibility.

Looks like this has been implemented now, but the comment wasn't collapsed.

vagrant/Vagrantfile
dcarley: 👌  could you add something to the commit message which explains that we need to generate/delete an SSH keypair because the Ubuntu AMI now that we're now using no longer contains the default Vagrant insecure key?

README.md
```diff
@@ -217,6 +217,24 @@ Stop the tunnel with: `make <env> stop-tunnel`
 ## Other useful commands
 Type `make` to get the list of all available commands.
 
+## Known caveats
+
+### Fly version
+
+Current mechanism of updating fly binary only checks if the date of your local
+binary is older than one on the server. If you have been using previous fly
+version recently, it won't get updated. In that case, delete the binary in /bin
```
dcarley, line 226: Does it not get updated by [`fly sync`](https://github.com/alphagov/paas-bootstrap/blob/4f42799af6c7a2c14188b701ea7bfbf852669baa/concourse/scripts/fly_sync_and_login.sh#L20-L21) (copied from and described in alphagov/paas-cf@3a2fd218ede057d2f3a32e12ef08c68018a889e1)?

mtekel: that's the 1st download. You can't do fly sync without fly. Or with fly that doesn't support it. Before fly sync, curl runs and it uses the described mechanism to tell if it should download the file or not.

mtekel:
Updated commit messages and implemented suggested improvements.

README.md
dcarley: Can you describe the scenario (actions and outcome) where it fails to work and the user would need to know about this section of the README?

mtekel: For example when you are upgrading from 0.7 (previous version). If you have downloaded it recently, curl won't update it and you get some cryptic error like `Resource type not found`.

paroxp: When the timestamp of the local binary is more recent than the one hosted on the server, you'll receive a `304 Not Modified` preventing to download a new file. The old (current) version of fly, doesn't support the `sync` functionality, preventing it from doing a self upgrade.

We could download the `fly` binary on each and every run, however this means there is this time and bandwidth "wasted".

This should possibly happen, only once - when upgrading from version < `1.0`. After this initial upgrade, the `fly` should either be able to sync itself or download if not existent.

saliceti:
Works fine 👍 
Could you squash commits 99f8975, 3a12b8c, a4e7138 and 2380a6f?

README.md
dcarley: > The old (current) version of fly, doesn't support the sync functionality, preventing it from doing a self upgrade.

Makes sense - thanks.

dcarley:
Well done 🎉

concourse/pipelines/destroy.yml
```diff
@@ -1,7 +1,7 @@
 ---
 resources:
```
alext, line 2: This is missing the `resource_types` block needed by all the s3 and semver resources. (Fix in #51)

concourse/pipelines/destroy.yml
```diff
@@ -81,7 +80,12 @@ jobs:
 
       - task: vpc-terraform-outputs-to-sh
         config:
-          image: docker:///ruby#2.2-slim
+          platform: linux
+          image_resource:
+            type: docker-image
+            source:
+              repository: governmentpaas/git-ssh
+              tag: "2.2-slim"
```
alext, line 88: I don't think this was meant to be updated to use the `git-ssh` container. There's no `2.2-slim` tag for that container. (Fix in #51)

PR #49, opened 2017-02-21
[#130020741] Concourse PW: don't generate nats password

mtekel:
## What

The "nats" password was used when concourse was deployed by bosh init. But
we deploy it bosh now and this password is not being used anywhere any more.
Clean it up.

## How to review

Verify that password is not being used anywhere. You can for example grep for it.

## Who can review

not @mtekel


keymon:
Seems to make sense. Merging

PR #50, opened 2017-02-24
[#119379405] Fix bootstrap pipelines

mtekel:
## What
In my enthusiasm I have forgotten to define outputs for mostly TF tasks.
That is, I let TF run directly on file that is in the input directory and
update it in place. That works fine, but if you try to 'put' that file
later, that is a different container and it doesn't have access to that
file. So you need to explicitly define it as an output. Fix the pipelines to
do so.

## How to review

Deploy a completely new environment from scratch. You can e.g. pick a name that's not used by anyone. All should work fine.

## Who can review

not @mtekel

### DIFF
--- a/concourse/pipelines/create.yml
+++ b/concourse/pipelines/create.yml
@@ -186,17 +186,21 @@ jobs:
           inputs:
             - name: paas-bootstrap
             - name: bucket-state
+          outputs:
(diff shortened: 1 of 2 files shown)


keymon:
It looks good. I will let it finish deploying my environment and if it works fine, merge.

PR #51, opened 2017-02-24
[#119379405] Fix destroy pipeline

alext:
## What

#48 Updated this repo to use Concourse 2.7. However that update has left the destroy pipeline in a broken state. This fixes it by adding in the `resource_types` block, and fixing container definition for the `extract-terraform-outputs` jobs.

## How to review

Test that the destroy pipeline works.

## Who can review

Anyone but myself.


paroxp (approved):
All good.

PR #52, opened 2017-02-28
[#130020741] Implement automation of credential rotation (continued)

mtekel:
## What

[Implement automation of credential rotation (continued)](https://www.pivotaltracker.com/story/show/130020741)

Add tasks to clean credentials we deem safe to rotate. 
Moved from paas-cf, see https://github.com/alphagov/paas-cf/pull/802

## How to review

Run the tasks to clear bosh and concourse credentials. Run the main pipeline to generate new creds and apply them. All should work fine.

## Who can review

not @mtekel


concourse/pipelines/create.yml
```diff
@@ -1184,7 +1204,8 @@ jobs:
               - -c
               - |
                 export CONCOURSE_URL="https://${CONCOURSE_HOSTNAME}.${SYSTEM_DNS_ZONE_NAME}"
-                export CONCOURSE_ATC_PASSWORD=$(./paas-bootstrap/concourse/scripts/val_from_yaml.rb secrets.concourse_atc_password concourse-secrets/concourse-secrets.yml)
+                export CONCOURSE_ATC_PASSWORD
+                CONCOURSE_ATC_PASSWORD=$(./paas-bootstrap/concourse/scripts/val_from_yaml.rb secrets.concourse_atc_password concourse-secrets/concourse-secrets.yml)
```
dcarley, line 1208: > Fix script so that it doesn't "swallow" exit code.

Could you explain in the commit message what the effect of this is before and after?

dcarley (changes requested):
Works. One small request about commit messages. Good to merge once updated.

concourse/pipelines/create.yml
mtekel: It was failing shellcheck. I have no idea how was this passing before. I've updated commit message.

alext: It might have been always failing shellcheck. I'm not sure that we've made the warnings generated by pipecleaner fail the build (we're not passing the `--fatal-warnings` flag to pipecleaner). We probably should now that the last one has been fixed.

mtekel: Added fatal warnings.

dcarley: Thanks 👍

dcarley: PS: I think that should have been a separate commit, but not important enough to hold up the merge.

dcarley (approved):

PR #54, opened 2017-03-01
[#140853603] Fix CONCOURSE_IP in ssh.sh script

saliceti:
## What
Story: [Fix ssh_concourse](https://www.pivotaltracker.com/story/show/140853603)

There was a typo that was really a problem because we don't check AWS credentials but this could bite us later on as we may do the same checks as in paas-cf.

We fixed it in https://github.com/alphagov/paas-cf/pull/811 so it makes sense to keep paas-bootstrap tidy as well.

## How to review
Try `ssh_concourse` on a non DEV environment

## Who can review
Not me


mtekel:
works, thanks for the fix

PR #56, opened 2017-03-17
Add boostrap concourse ssh target

mtekel:
## What

Add makefile target to ssh to bootstrap concourse. Display bootstrap concourse IP in showenv.

## How to review

Deploy your bootstrap server if you don't have one already. Run `make dev ssh_bootstrap_concourse`. 
Run `make dev showenv`. Check that the it also contains bootstrap concourse IP.

## Who can review

not @mtekel

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,6 +147,9 @@ showenv: ## Display environment information
 ssh_concourse: check-env-vars ## SSH to the concourse server
 	@./concourse/scripts/ssh.sh
 
+ssh_bootstrap_concourse: check-env-vars ## SSH to the bootstrap concourse server
+	@cd vagrant ; vagrant ssh -- -i ../${VAGRANT_SSH_KEY_NAME}
+
 tunnel: check-env-vars ## SSH tunnel to internal IPs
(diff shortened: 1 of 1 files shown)


alext:
I don't think this change is correct. I expect these two cases to operate on the concourse that this repo creates (ie the deployer/build concourse).

If we want to add tasks to access the temporary bootstrap concourse, then these should be added separately IMO.

mtekel:
updated

combor:
@alext I agree that we we should have separate tasks for ssh-ing to deployer and bootstrap Concourse. What I'm not sure about is where to put them but IMO we should avoid code duplication. Now we've got `ssh_concourse` target in `paas-bootstrap` and `paas-cf` but not in `paas-release-ci`. So I would got for either:
- centralise ssh_concourse in `paas-bootstrap` or
- implement `ssh_concourse` in `paas-release-ci` and remove from `paas-bootstrap`

mtekel:
Agreed, but looks a bit out of scope of this PR. We have a story for that kind of things: https://www.pivotaltracker.com/story/show/137365155

PR #59, opened 2017-04-21
[#141315433] Add deploy_env tag to BOSH RDS

dcarley:
## What

This tag will propagate through to the metrics that we get from the RDS
integration in DataDog. We can then use it to restrict monitors to a single
environment, rather than every environment in the AWS account (dev and CI
host more than one).

## How to review

You can't test the metrics because the DataDog RDS integration isn't enabled in dev 🐼 

So I suggest:

1. Sanity check the code.
1. Optionally deploy to your dev environment to double check that the tag is applied to the instance.
1. Deploy to CI and observe metrics appearing in [this graph](https://app.datadoghq.com/metric/explorer?live=true&page=0&is_auto=false&from_ts=1492782789412&to_ts=1492786389412&tile_size=m&exp_metric=aws.rds.cpucredit_balance&exp_scope=deploy_env%3Amaster&exp_group=dbinstanceidentifier&exp_agg=avg&exp_row_type=metric)
1. Merge
1. Deploy to all other environments (sorry!)

## Who can review

Not @dcarley

### DIFF
--- a/terraform/bosh/rds.tf
+++ b/terraform/bosh/rds.tf
@@ -57,6 +57,7 @@ resource "aws_db_instance" "bosh" {
   vpc_security_group_ids = ["${aws_security_group.bosh_rds.id}"]
 
   tags {
-    Name = "${var.env}-bosh"
+    Name       = "${var.env}-bosh"
+    deploy_env = "${var.env}"
   }
 }


saliceti:
LGTM

PR #68, opened 2017-06-07
README, Add note regarding DEPLOY_ENV

LeePorte:
## What

A few of us have been caught out by using the same `DEPLOY_ENV` for both
build and deployer concourse environments. This adds a note to inform
the reader to use a different one to prevent resource allocation
conflicts.

## How to review

Code review.

## Who can review

Not me

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,8 @@ these are defined in the repo [aws-account-wide-terraform][]
 $ export DEPLOY_ENV=environment-name
 ```
 
+It is important that you do not use the same `DEPLOY_ENV` for both build and deployer concourse environments, this is to avoid conflicts in resource allocation.
+
 ### Deploy
 
 Create Concourse Lite with `make`. There are targets to select the target AWS account, and to select the profiles to apply. For instance for a DEV build concourse bootstrap:


README.md
```diff
@@ -66,6 +66,8 @@ these are defined in the repo [aws-account-wide-terraform][]
 $ export DEPLOY_ENV=environment-name
 ```
 
+It is important that you do not use the same `DEPLOY_ENV` for both build and deployer concourse environments, this is to avoid conflicts in resource allocation.
```
paroxp, line 69: I'm wondering if replacing `important` with `recommended` would be a better fit here.

paroxp (approved):
I'm happy with the change as is. I still consider the commend I've left. Happy to push it through if you feel otherwise.

README.md
LeePorte: I'm indifferent to changing them. I guess if we go with important then we're less likely to see issues.

dcarley: Recommended suggests that there are reasons why you my want to do this in purpose, which I don't think is true. Important is good 👍

dcarley:
I'd like to find a technical solution to this in the future. But this is a good improvement in the meantime, thanks!

PR #70, opened 2017-06-15
[#139001857] Allow Bosh RDS instance to auto_minor_version_upgrade

alext:
## What

So that we automatically get upgraded to newer point releases when there
are security issues etc.

## How to review

* Run the bootstrap pipeline.
* Observe that the Bosh RDS instance gets auto_minot_version_upgrade set to true.
* Observe that terraform doesn't attempt to change the engine version.

## Who can review

Not me.


combor:
LGTM. Merging.

PR #71, opened 2017-06-21
[#147124125] Increase max number of retries

paroxp:
## What
We're implementing a functionality that'll kick-off the
dev `create-cloudfoundry` pipeline every morning. Starting many
environments at roughly the same time, will cause increased amount of
calls made towards AWS.

Increasing the number of max retries, should help us a little with the
potential issue of being caught out by the limit established upstream.

It is set to 16, as previously (a while ago), we had it set to 8. We
have more people on the team and 16 felt like a safe bet to play with.

I have ran the bootstrap on my `dev` environment, which didn't cause any
issues with regards to the change.

## How to review

I don't think it is something easy to review.

- Check if the value is correct
- Run `make dev bootstrap`
- Be aware of its existence and monitor the behaviour over time

### DIFF
--- a/manifests/bosh-manifest/bosh-manifest.yml
+++ b/manifests/bosh-manifest/bosh-manifest.yml
@@ -13,6 +13,7 @@ meta:
     default_security_groups:
     - (( grab terraform_outputs.bosh_managed_security_group ))
     region: (( grab terraform_outputs.region ))
+    max_retries: 16
 
   bosh_private_ip: (( grab terraform_outputs.microbosh_static_private_ip ))
   bosh_public_ip: (( grab terraform_outputs.microbosh_static_public_ip ))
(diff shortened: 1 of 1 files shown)


dcarley:
I'm going to look at this as part of deploying a new stemcell to save us running paas-bootstrap twice. I have a couple of requests:

- could you add a bit more detail to the first sentence of the commit message to say that we're planning to make the `create-cloudfoundry` pipeline start automatically every morning for development environments?
- could you note whether you've tested this in a development environment yourself? You don't necessarily need to test whether it reduces the chance of hitting the rate limit, but it would be good to know whether the config was syntactically correct and deployed successfully

dcarley:
> I have ran the bootstrap on my `dev` environment, which didn't cause any issues with regards to the change.

👍  I'm going to code review only and deploy with #72

PR #74, opened 2017-06-29
[#148059799] Update datadog-agent boshrelease

paroxp:
## What

We've applied some logrotate fixes to the boshrelease, which will no
longer prevent logrotate from rotating the logs on datadog-agent.

## How to review

- Code review
- Run the bootstrap from this branch - inb4 Dan: I have run that myself :)
- Experience no issues
- For bonus points: Check if logrotate does warn about permissions

## Before merging

This PR highly depends on the [`datadog-agent`](https://github.com/alphagov/paas-datadog-agent-boshrelease/pull/11) and [`collect`](alphagov/paas-collectd-boshrelease#3) being merged first. The changes here will need to be upgraded with new correct versions.

### DIFF
--- a/manifests/runtime-config/addons-meta/datadog-agent.yml
+++ b/manifests/runtime-config/addons-meta/datadog-agent.yml
@@ -1,9 +1,9 @@
 ---
 releases:
   - name: datadog-agent
-    version: 0.1.4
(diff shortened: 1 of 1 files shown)


alext:
Upon further investigation, we think this change is unnecessary, and a side-effect of how we were running logrotate while testing (see https://github.com/onemedical/datadog-agent-boshrelease/pull/17 for details).

I'm therefore closing this.

PR #75, opened 2017-07-04
[#137364635] - Add self updating pipeline to concourses

bleach:
## What

Previously we applied any changes from this repo to persistent environments using a bootstrap concourse lite spun up for the purpose.

In order to reduce the amount of human time required to make deployment changes, we will apply those changes directly from the build or deployer concourse.

When this PR is merged:
- Bootstrap concourse will be used to create and destroy build/deployer concourse and their BOSH
- Changes will be applied to persistent environments by logging into the build or deployer concourse and starting a pipeline run

## How to review

This change is dependent on [IAM changes in a corresponding PR](https://github.digital.cabinet-office.gov.uk/government-paas/aws-account-wide-terraform/pull/96) which should be merged first. 

The most important path to test is how persistent environments will behave when we apply these changes.

### Deployer concourse

- Create a deployer concourse from master and then destroy the bootstrap concourse OR use your existing one.
- Spin up a bootstrap concourse from this branch: `DEPLOY_ENV=<env> BRANCH=add-self-update-pipeline make dev deployer-concourse bootstrap`
- Run the [create-bosh-concourse](pipeline) which should add a new pipeline to your deployer concourse.
- Run the same pipeline on your deployer concourse

### Build concourse

- Create a build concourse from master and then destroy the bootstrap concourse
- Spin up a bootstrap concourse from this branch: `DEPLOY_ENV=<env> BRANCH=add-self-update-pipeline make dev build-concourse bootstrap`
- Run the [create-bosh-concourse](pipeline) which should add a new pipeline to your build concourse.
- Run the same pipeline on your build concourse

### Merging & rolling out

- Check that the [account-wide-terraform PR](https://github.digital.cabinet-office.gov.uk/government-paas/aws-account-wide-terraform/pull/96) has been merged & applied
- [Merge using a signed commit](https://government-paas-team-manual.readthedocs.io/en/latest/team/working_practices/#merging-pull-requests)
- In each of CI master, CI build, staging, prod:
  - create a bootstrap concourse and run the `create-bosh-concourse` pipeline from it
  - run the same pipeline on the persistent concourse

## Who can review

Not @bleach or @henrytk

### DIFF
--- a/concourse/pipelines/create-bosh-concourse.yml
+++ b/concourse/pipelines/create-bosh-concourse.yml
@@ -1139,6 +1139,7 @@ jobs:
 
   - name: concourse-deploy
     serial: true
+    interruptible: true
     plan:
       - aggregate:
         - get: paas-bootstrap
--- a/concourse/pipelines/create-bosh-concourse.yml
+++ b/concourse/pipelines/create-bosh-concourse.yml
@@ -13,7 +13,7 @@ groups:
       - generate-concourse-config
       - concourse-deploy
       - post-deploy
-      - expunge-vagrant
+      - expunge-concourse
 
   - name: credentials
     jobs:
@@ -239,7 +239,6 @@ jobs:
             SELF_UPDATE_PIPELINE: {{self_update_pipeline}}
             TARGET_CONCOURSE: {{target_concourse}}
             CONCOURSE_TYPE: {{concourse_type}}
-            VAGRANT_IP: {{vagrant_ip}}
           run:
             path: ./paas-bootstrap/concourse/scripts/self-update-pipeline.sh
 
@@ -330,7 +329,6 @@ jobs:
           outputs:
           - name: updated-vpc-tfstate
           params:
-            VAGRANT_IP: {{vagrant_ip}}
             TF_VAR_env: {{deploy_env}}
             AWS_DEFAULT_REGION: {{aws_region}}
           run:
@@ -339,7 +337,8 @@ jobs:
             - -e
             - -c
             - |
-              terraform_params=${VAGRANT_IP:+-var vagrant_cidr=$VAGRANT_IP/32}
+              CONCOURSE_EGRESS_IP=$(wget -q -O - http://169.254.169.254/latest/meta-data/public-ipv4)
+              terraform_params="-var concourse_egress_cidr=${CONCOURSE_EGRESS_IP}/32"
               terraform apply ${terraform_params} -var-file=paas-bootstrap/terraform/{{aws_account}}.tfvars \
                 -state=vpc-tfstate/vpc.tfstate \
                 -state-out=updated-vpc-tfstate/vpc.tfstate \
@@ -604,7 +603,6 @@ jobs:
           params:
             DEPLOY_ENV: {{deploy_env}}
             AWS_DEFAULT_REGION: {{aws_region}}
-            VAGRANT_IP: {{vagrant_ip}}
             TF_VAR_system_dns_zone_name: {{system_dns_zone_name}}
             TF_VAR_bosh_az: {{bosh_az}}
             TF_VAR_bosh_fqdn: {{bosh_fqdn}}
@@ -620,7 +618,8 @@ jobs:
 
                 cp ssh-public-key/id_rsa.pub paas-bootstrap/terraform/bosh
                 cp bosh-ssh-public-key/bosh_id_rsa.pub paas-bootstrap/terraform/bosh
-                terraform_params=${VAGRANT_IP:+-var vagrant_cidr=$VAGRANT_IP/32}
+                CONCOURSE_EGRESS_IP=$(wget -q -O - http://169.254.169.254/latest/meta-data/public-ipv4)
+                terraform_params="-var concourse_egress_cidr=${CONCOURSE_EGRESS_IP}/32"
                 terraform apply ${terraform_params} -var env={{deploy_env}} -var-file=paas-bootstrap/terraform/{{aws_account}}.tfvars \
                   -state=bosh-tfstate/bosh.tfstate -state-out=updated-bosh-tfstate/bosh.tfstate paas-bootstrap/terraform/bosh
         ensure:
@@ -727,7 +726,6 @@ jobs:
               ./terraform-outputs/vpc.terraform-outputs.yml
             BOSH_FQDN: {{bosh_fqdn}}
             BOSH_FQDN_EXTERNAL: {{bosh_fqdn_external}}
-            BOSH_LOGIN_HOST: {{bosh_login_host}}
             BOSH_INSTANCE_PROFILE: {{bosh_instance_profile}}
           run:
             path: sh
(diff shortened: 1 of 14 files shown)


concourse/pipelines/destroy-bosh-concourse.yml
```diff
@@ -84,10 +84,41 @@ resources:
...
+            AWS_ACCOUNT: {{aws_account}}
+            SKIP_COMMIT_VERIFICATION: {{skip_commit_verification}}
+            SELF_UPDATE_PIPELINE: {{self_update_pipeline}}
+            TARGET_CONCOURSE: {{target_concourse}}
+            CONCOURSE_TYPE: {{concourse_type}}
+            VAGRANT_IP: {{vagrant_ip}}
+          run:
+            path: ./paas-bootstrap/concourse/scripts/self-update-pipeline.sh
+      - put: pipeline-trigger
+        params: {bump: patch}
```
alext: Adding this here makes sense, but we need to also remove the bump from the end of the `enable-bosh-access` job (line 189)

concourse/scripts/environment.sh
```diff
@@ -17,15 +17,40 @@ fi
...
+        user_id=$(aws sts get-caller-identity | awk '$1 ~ /UserId/ {print $2}')
+        CONCOURSE_ATC_PASSWORD=$(hashed_password "${user_id}")
+    fi
+    BOSH_LOGIN_HOST=${BOSH_FQDN_EXTERNAL}
+    ;;
+  deployer-concourse|build-concourse)
+    if [ "$TARGET_CONCOURSE" == 'deployer-concourse' ]; then
+      CONCOURSE_URL="${CONCOURSE_URL:-"https://deployer.${SYSTEM_DNS_ZONE_NAME}"}"
+    else
+      CONCOURSE_URL="${CONCOURSE_URL:-"https://concourse.${SYSTEM_DNS_ZONE_NAME}"}"
```
alext: This logic is already captured by the `CONCOURSE_HOSTNAME` variable set in the `Makefile`. Would it be possible to re-use that instead of having to repeat the information here.

concourse/scripts/self-update-pipeline.sh
```diff
@@ -0,0 +1,30 @@
...
+# Optional variables:
+# - BRANCH
+
+set -u
+set -e
+
+if [ ! -d "./paas-bootstrap" ]; then
+  echo "Resource paas-bootstrap must be checked out"
+  exit 1
+fi
```
alext: This check seems redundant given this script is located within the paas-bootstrap repo.

concourse/scripts/self-update-pipeline.sh
```diff
@@ -0,0 +1,30 @@
...
+
+if [ "${TARGET_CONCOURSE}" == "bootstrap" ]; then
+  echo "Bootstrap Concourse should not self-update, as this requires access to \`aws sts get-caller-identity\`. Skipping."
+  exit 0
+fi
+
+if [ "${SELF_UPDATE_PIPELINE}" != "true" ]; then
```
alext: Can we reverse this conditional. As it stands, it's effectively an `unless...else...` clause, which takes a moment to parse.

alext (commented):
I've added a few comments from a code-review. I've not tested this running yet. I'll do that now.

concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -1229,20 +1266,48 @@ jobs:
...
+          inputs:
+            - name: paas-bootstrap
+          params:
+            DEPLOY_ENV: {{deploy_env}}
+            BRANCH: {{branch_name}}
+            AWS_ACCOUNT: {{aws_account}}
+            SKIP_COMMIT_VERIFICATION: {{skip_commit_verification}}
+            SELF_UPDATE_PIPELINE: {{self_update_pipeline}}
+            CONCOURSE_TYPE: {{concourse_type}}
+            VAGRANT_IP: {{vagrant_ip}}
```
alext: I don't think we want to pass this through here. I think we need to explicitly unset it.

As it is at the moment, when the pipeline runs on the deployer-concourse, it re-adds this IP to the security groups for the duration of the pipeline run. The bootstrap concourse instance that this IP belongs to will likely be long-gone by this point.

alext: I've run into another issue related to this:

I attempted a stemcell bump for bosh running with the deployer-concourse, and it timed out attempting to connect to the bosh agent on the bosh VM.

It's tricky to add the concourse IP to the bosh security groups because it's not necessarily known at the point that we run bosh-terraform (unless we were to hardcode it in the bosh-manifest and in terraform).

An alternative would be to rework this so that it contains the IP of the concourse that's running the pipeline (the variable will need renaming as well).

bleach: What I'm testing at the moment:
- add an ingress rule allowing the `bosh_api_client` SG to talk to port 6868 on the internal BOSH IP
- switch to using the internal IP of the bosh director VM everywhere. Currently using a hardcoded commit, but will have to add logic to ensure that bootstrap concourse continues to use the public IP

do you foresee problems with this approach?

alext: > switch to using the internal IP of the bosh director VM everywhere. Currently using a hardcoded commit, but will have to add logic to ensure that bootstrap concourse continues to use the public IP

I'm not sure how this can work because the bootstrap concourse runs outside the VPC (it creates the VPC so can't run within it). This is the reason we're using the external address in the first place.

I think we'll either need to make both concourse's use the external IP when deploying bosh, or we'll need to extend the use of `BOSH_LOGIN_HOST` to bosh-init (and the bosh manifest) as well.

bleach: > or we'll need to extend the use of BOSH_LOGIN_HOST to bosh-init (and the bosh manifest) as well

I think this is my preferred approach

concourse/pipelines/destroy-bosh-concourse.yml
bleach: Done

concourse/scripts/environment.sh
bleach: I've reused it and also in another place in the manifests where we were generating it again.

I left `TARGET_CONCOURSE` alone though.

concourse/scripts/self-update-pipeline.sh
bleach: Removed. Looks like it was originally copied from the [paas-cf implementation](https://github.com/alphagov/paas-cf/blob/4200fb86f90666531956b6043690094e283f530a/concourse/scripts/self-update-pipeline.sh#L15). I'll raise it there too.

concourse/pipelines/create-bosh-concourse.yml
bleach: I've switched to using `BOSH_LOGIN_HOST` and BOSH director stemcell changes now work from my deployer concourse.

concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -1198,7 +1237,6 @@ jobs:
               - -e
               - -c
               - |
-                export CONCOURSE_URL="https://${CONCOURSE_HOSTNAME}.${SYSTEM_DNS_ZONE_NAME}"
```
alext: This has to remain as it was. This job needs to always connect to the deployed concourse. With this change, it fails when run on the bootstrap concourse as it tried to connect to `localhost:8080` which fails.

bleach: Restored this

concourse/scripts/self-update-pipeline.sh
bleach: Done

concourse/pipelines/create-bosh-concourse.yml
bleach: I've removed `VAGRANT_IP` from that job.

bleach:
I think this is ready for code review again. I'm still testing that the pipelines work from bootstrap, build and deployer concourses.

After you've code reviewed I think it would make sense to squash all the commits except a255804 (switching to accessing BOSH over its public IP) into the original commits. That one has some useful context on why we did it.

alext:
This all looks good now.

We can merge this once it's been squashed/rebased, and the role_policy_attachment limit issue is resolved in CI.

bleach:
I fixed a bug in the destroy pipeline.

I'll squash / rebase on Monday, as I've run out of time today.

bleach:
The commits have been squashed now. We're blocked on the associated account-wide-terraform PR which itself is blocked on an IAM limits raise.

PR #78, opened 2017-07-10
README: Make the description more approachable

46bit:
The existing README is written in technical language that assumes a lot 
of PaaS-specific knowledge. Setting it up for myself involved a lot of 
questions as to its meaning. 

## What

This adds a one-line explanation that is understandable with basic knowledge of Vagrant. In particular my questions were:

* What does it mean to be outside an environment? (I didn't ask this one because it didn't seem crucial.)
* Is a VPC the same as an environment?
* Where does the Vagrant instance run? (At first I'd assumed it ran locally but then it was talking about an AWS plugin.)

I think this description of Concourse Lite gives enough context at a basic level for people to figure these questions out themselves. Explaining that Vagrant is using AWS made the instructions lengthier and broke up the this-then-that layout they have at present.

I'm very much open to feedback on how to improve this or change further.

## How to review

Examine whether the changes seem helpful, don't omit important technical detail for the experienced, (and whether anything else should be done.)

## Who can review

Anyone except @LeePorte and I.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,11 +11,15 @@ any kind of deployment environment.
 
 ## Concourse Lite
 
-This runs outside an environment and is responsible for creating and
-destroying a VPC containing Bosh and Concourse.
+Concourse Lite is responsible for bootstrapping a full Bosh and Concourse 
+environment ([paas-cf][]). First it creates a Vagrant instance running on AWS.
+Second this Vagrant instance is used to provision the full environment.
+
 You don't need to keep this running once Concourse is deployed,
 and you can create it again when Concourse needs to be modified
 or destroyed.
+
+[paas-cf]: https://github.com/alphagov/paas-cf
 
 ### Prerequisites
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,12 +11,16 @@ any kind of deployment environment.
 
 ## Concourse Lite
 
-This runs outside an environment and is responsible for creating and
-destroying a VPC containing Bosh and Concourse.
+Concourse Lite is responsible for bootstrapping a full Bosh and Concourse 
+environment ([paas-cf][]). First it creates a Vagrant instance running on AWS.
```
alext, line 15: This link isn't quite correct in this context. paas-cf uses this Bosh and Concourse to deploy CloudFoundry.

README.md
```diff
@@ -11,12 +11,16 @@ any kind of deployment environment.
...
+Second this Vagrant instance is used to provision the full environment.
```
alext, line 16: This instance isn't used for anything other than provisioning the Bosh and Concourse. Once that step is completed, this is no longer used.

46bit:
@alext I could also add an explanation of what the build and deployer concourses are (CI-only environment and CloudFoundry respectively) further down. It's probably best if I revisit this after tomorrow's architecture talk when I'll have a better understanding.

46bit:
I think it would help to document how the pipeline setup works for onboarding, but I don't feel that I'm comfortable enough with the architecture yet to do it. Closing for now.

PR #80, opened 2017-07-24
[#137364635] Fix pipelines.sh when SKIP_COMMIT_VERIFICATION unset.

alext:
## What

This was erroring when this variable isn't set. It's only typically set
in dev environments.

On line 47 of this script we're using ${SKIP_COMMIT_VERIFICATION:-}, but
this hadn't been repeated in the generate_vars_file block, leading to
the error:
```
.../paas-bootstrap/concourse/scripts/pipelines.sh:
line 11: SKIP_COMMIT_VERIFICATION: unbound variable
```

## How to review

* Delete the `$(eval export SKIP_COMMIT_VERIFICATION=true)` line from `dev` target in the `Makefile` locally.
* Create a bootstrap concourse (`make dev deployer-concourse bootstrap`)
* Verify that it doesn't error when pushing the pipelines to this concourse.

## Who can review

Not me.


bleach:
LGTM and works in my dev env

PR #82, opened 2017-07-24
Use gem to merge and sign PRs

alext:
## What

We've extracted this into a gem, so there's no longer a need to maintain
the script in this repo.

This reverts some of the changes made in 309302f when these scripts were
added.

See also https://github.com/alphagov/paas-cf/pull/987 for the same change in paas-cf.

## How to review

Code review. The gem itself has already been reviewed here: alphagov/paas-github_merge_sign#1

Optionally, you could use this branch to merge a PR to verify it all works.

## Who can review

Not me.


dcarley:
I used this PR to merge itself (because the code executes before the branch is switched) 😁

PR #89, opened 2017-09-12
[#150512756] IPv4 Keyservers

alext:
## What

We're experiencing an intermittent problem since we've upgraded
Concourse and garden-runc where gpg will intermittently return an error:
```
gpg: keyserver receive failed: Address family not supported by protocol
```
Switching to an IPv4-only pool works around this, and will unblock the
pipeline while we investigate further.

## How to review

Code review is probably enough if you've already reviewed alphagov/paas-cf#1035

## Who can review

Not me.


dcarley (approved):
👍  I can't sign though, so feel free to merge yourself.

PR #98, opened 2017-10-30
[#152067205] Improve shellcheck linting for pipelines

alext:
## What

This is the companion to https://github.com/alphagov/paas-cf/pull/1083 and re-enables some shellcheck checks for the concourse pipelines.

## How to review

This should probably be reviewed after https://github.com/alphagov/paas-cf/pull/1083 as it pulls in the pipecleaner changes from that PR.

Code review is probably enough (may be easier commit-by-commit).

You could run the pipelines from this branch to test if you deem it necessary.

## Who can review

Not me.

### DIFF
--- a/concourse/pipelines/create-bosh-concourse.yml
+++ b/concourse/pipelines/create-bosh-concourse.yml
@@ -187,7 +187,7 @@ jobs:
             - -e
             - |
               cd bucket-state
-              aws s3 cp s3://((state_bucket))/bucket.tfstate . || true
+              aws s3 cp "s3://((state_bucket))/bucket.tfstate" . || true
               ls -l
 
       - task: create-init-bucket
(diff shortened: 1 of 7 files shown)


concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -745,6 +749,7 @@ jobs:
                                         ./paas-bootstrap/manifests/runtime-config/addons-meta/collectd.yml"
                 fi
 
+                # shellcheck disable=SC2086
```
camelpunch, line 752: Why are we disabling the double quote check here? There are a few other disables on this file too.

alext: because in this instance, we want this to be split into multiple args to the `build_manifest.sh` script. This variable is set to the list of manifest files to be merged into a single output.

camelpunch: Makes sense, thanks.

camelpunch (approved):

PR #100, opened 2017-11-10
[#150711834] Add scripting and security groups for SSHing directly to Bosh

henrytk:
## What

https://www.pivotaltracker.com/story/show/150711834

Add the security group for accessing Bosh via SSH directly and the script for using a container to operate the Bosh Director.

## How to review

Review in conjunction with https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/101, which _**must**_ be merged first.

* Deploy a Bosh from this branch and check it gets the usual `<env>-bosh` security and the new `office-access-ssh` one we have added.
* Run `DEPLOY_ENV=<deploy_env> make bosh-cli` with your AWS credentials loaded.
  * If the above pull request above has been merged and Docker Hub has built the image then you can remove the temporary commit and run against the latest image. Otherwise, keep the temporary commit and it will point to the dev version of the image.
* The Bosh checks you need to run to confirm it works are detailed in the description of the other pull request: https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/101
* Check the updated IP whitelist is correct.

### Before merging

* the above pull request needs to be merged and Docker Hub needs to build the image from master.
* the temporary commit from this branch needs to be removed.

### After merging

This needs to be applied to persistent environments. The `create-bosh-concourse` pipeline is present on the persistent environment concourses, so you can do the manual clicky clicky.

## Who can review

Anyone but me or @camelpunch

### DIFF
--- a/scripts/bosh-cli.sh
+++ b/scripts/bosh-cli.sh
@@ -20,4 +20,4 @@ docker run \
     --env "BOSH_ID_RSA" \
     --env "BOSH_IP" \
     --env "BOSH_ADMIN_PASSWORD" \
-    governmentpaas/bosh-shell
+    governmentpaas/bosh-shell:f804160a7e6f455ef206b56eab25063908703dd5
--- a/scripts/bosh-cli.sh
+++ b/scripts/bosh-cli.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+set -eu
+
+BOSH_ID_RSA="$(aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/bosh_id_rsa" - | base64)"
+export BOSH_ID_RSA
+
+BOSH_IP=$(aws ec2 describe-instances \
+    --filters "Name=key-name,Values=${DEPLOY_ENV}_bosh_ssh_key_pair" \
+    --query 'Reservations[].Instances[].PublicIpAddress' --output text)
+export BOSH_IP
+
+BOSH_ADMIN_PASSWORD=$(aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/bosh-secrets.yml" - | \
+    ruby -ryaml -e 'print YAML.load(STDIN)["secrets"]["bosh_admin_password"]')
+export BOSH_ADMIN_PASSWORD
+
+docker run \
+    -it \
+    --rm \
+    --env "BOSH_ID_RSA" \
+    --env "BOSH_IP" \
+    --env "BOSH_ADMIN_PASSWORD" \
+    governmentpaas/bosh-shell
--- a/scripts/ssh_bosh.sh
+++ b/scripts/ssh_bosh.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+set -euo pipefail
+
+BOSH_KEY=/tmp/bosh_id_rsa.$RANDOM
+BOSH_IP=$(aws ec2 describe-instances \
+    --filters "Name=key-name,Values=${DEPLOY_ENV}_bosh_ssh_key_pair" \
+    --query 'Reservations[].Instances[].PublicIpAddress' --output text)
+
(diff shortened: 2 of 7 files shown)


scripts/bosh-cli.sh
```diff
@@ -0,0 +1,20 @@
...
+
+BOSH_ADMIN_PASSWORD=$(aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/bosh-secrets.yml" - | \
+    ruby -ryaml -e 'print YAML.load(STDIN)["secrets"]["bosh_admin_password"]')
+
+docker run \
+    -it \
+    --rm \
+    --env "BOSH_ID_RSA=$BOSH_ID_RSA" \
+    --env "BOSH_IP=$BOSH_IP" \
+    --env "BOSH_ADMIN_PASSWORD=$BOSH_ADMIN_PASSWORD" \
```
alext: This exposes both the password and the SSH private key in the `ps` output on the host. Is there another way these could be passed into the container? Maybe we should look at using `--env-file` instead.

alext (changes requested):
This approach looks to work well. I've spotted a couple of issues though that we should address...

scripts/ssh_bosh.sh
```diff
@@ -0,0 +1,23 @@
...
+
+trap 'rm -f ${BOSH_KEY}' EXIT
+
+aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/bosh_id_rsa" ${BOSH_KEY} && chmod 400 ${BOSH_KEY}
+
+aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/bosh-secrets.yml" - | \
+  ruby -ryaml -e 'puts "Sudo password is " + YAML.load(STDIN)["secrets"]["bosh_vcap_password_orig"]'
+echo
+
+# shellcheck disable=SC2086
```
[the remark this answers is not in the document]
henrytk: This is an interesting one: if we quote it we get a different Shellcheck problem warning us it will be expanded on the remote side.

scripts/bosh-cli.sh
henrytk: Our main reason for choosing a container and environment variables was to avoid complications with cleaning up files, such as:

* process traps not executing because the signal can't be caught (`kill -9`, for example)
* system crashes
* scripting errors

If the chance of accidentally exposing `ps` output (or having it exploited) is greater than the risks of the above points then we can use `--env-file`, but I can't see how it is.

scripts/ssh_bosh.sh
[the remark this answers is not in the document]
alext: Hmmm... I think it would be better to quote it and disable `SC2029` as it's a more specific check. I'm also not convinced that the check is quite correct in this case as  it only fires when using the `user@host` syntax. It doesn't fire when just specifying a host, even with a variable in there.

scripts/bosh-cli.sh
alext: We already have the pattern of exit traps cleaning up secrets files in several places in our codebase (eg the `ssh_bosh.sh` script in this PR), so I don't think it's a problem to use the same pattern here. I think that's preferable to having this available in the `ps` output.

henrytk: As discussed offline I have pushed up a commit which exports the environment variables and lets Docker propagate them into the container. This avoids using `--env-file` and stops the secrets showing via `ps` to non-process owners.

scripts/ssh_bosh.sh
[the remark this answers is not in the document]
henrytk: Done

alext:
I've squashed the fixes.

I've also amended the container version pinning to use the commit-based tag now that alphagov/paas-docker-cloudfoundry-tools#101 has been merged.

alext (approved):

PR #101, opened 2017-11-15
[#152143894] Use oAuth for logging into concourse

LeePorte:
## What

This PR introduces Github oAuth to log into Concourse. An additional make target has been created to facilitate uploading of API credentials for authentication of users.

`# shellcheck disable=SC2046,SC2116` is used for the team creation command due to the shellcheck incorrectly reporting quoting errors and an unnecessary echo. 

## How to review

1. Create an oAuth application at https://github.com/settings/applications/new
1. Authorization callback URL should be `https://deployer.<your env>.dev.cloudpipeline.digital/auth/github/callback`
1. Store the client ID and Client secret in your personal pass store under the following locations `github.com/concourse/dev/client_id` and `github.com/concourse/dev/client_secret`
1. Checkout this branch and run `make dev upload-github-oauth GITHUB_PASSWORD_STORE_DIR=<your pass dir> DEPLOY_ENV= <your Env>`
1. Push the pipeline with the following command `BRANCH=$(git rev-parse --abbrev-ref HEAD) make dev pipelines CONCOURSE_TYPE=deployer-concourse CONCOURSE_HOSTNAME=deployer BOSH_INSTANCE_PROFILE=bosh-director-cf CONCOURSE_INSTANCE_TYPE=m4.xlarge CONCOURSE_INSTANCE_PROFILE=deployer-concourse ENABLE_COLLECTD_ADDON=true ENABLE_SYSLOG_ADDON=true ENABLE_GITHUB=true`
1. Run the `create-bosh-concourse` pipeline
1. once it has completed, login to team main using the login with Github button.

## Merging

Github API credentials are already in the persistent environment state buckets. Once merge to master the pipeline needs to be run in each environment.

## Who can review

Not @LeePorte or @46bit

### DIFF
--- a/concourse/pipelines/create-bosh-concourse.yml
+++ b/concourse/pipelines/create-bosh-concourse.yml
@@ -160,6 +160,13 @@ resources:
       region_name: ((aws_region))
       versioned_file: concourse-manifest.yml
 
+  - name: github-oauth-secrets
+    type: s3-iam
+    source:
+      bucket: ((state_bucket))
+      region_name: ((aws_region))
+      versioned_file: github-oauth-secrets.yml
+
 jobs:
   - name: init-bucket
     serial: true
@@ -279,6 +286,7 @@ jobs:
               paas-bootstrap/concourse/scripts/s3init.sh "((state_bucket))" id_rsa.pub paas-bootstrap/concourse/init_files/zero_bytes
               paas-bootstrap/concourse/scripts/s3init.sh "((state_bucket))" concourse.tfstate paas-bootstrap/concourse/init_files/terraform.tfstate.tpl
               paas-bootstrap/concourse/scripts/s3init.sh "((state_bucket))" concourse-secrets.yml paas-bootstrap/concourse/init_files/zero_bytes
+              paas-bootstrap/concourse/scripts/s3init.sh "((state_bucket))" github-oauth-secrets.yml paas-bootstrap/concourse/init_files/zero_bytes
 
   - name: vpc
     serial: true
@@ -1153,6 +1161,7 @@ jobs:
         - get: concourse-manifest
           passed: ['generate-concourse-config']
         - get: concourse-secrets
+        - get: github-oauth-secrets
 
       - task: get-and-upload-stemcell
         config:
@@ -1221,6 +1230,7 @@ jobs:
             - name: paas-bootstrap
             - name: concourse-manifest
             - name: concourse-secrets
+            - name: github-oauth-secrets
           params:
             CONCOURSE_HOSTNAME: ((concourse_hostname))
             SYSTEM_DNS_ZONE_NAME: ((system_dns_zone_name))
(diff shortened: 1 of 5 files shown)


concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -1221,13 +1230,17 @@ jobs:
...
           params:
             CONCOURSE_HOSTNAME: ((concourse_hostname))
             SYSTEM_DNS_ZONE_NAME: ((system_dns_zone_name))
             CONCOURSE_ATC_USER: admin
             FLY_TEAM: ((aws_account))
             FLY_TARGET: ((deploy_env))
             FLY_CMD: ./paas-bootstrap/bin/fly
+            ENABLE_GITHUB: ((enable_github))
+            GITHUB_CLIENT_ID: ((github_client_id))
+            GITHUB_CLIENT_SECRET: ((github_client_secret))
```
alext, line 1243: Why are we passing these in here if the task is also pulling them out of the file from the state bucket? This seems to duplicate the same thing.

I don't think these 2 pipeline variables are necessary with the current implementation.

concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -1237,6 +1250,38 @@ jobs:
...
+                  else
+                    export GITHUB_USERS="${GITHUB_PROD_USERS}"
+                  fi
+
+                  ./paas-bootstrap/concourse/scripts/fly_sync_and_login.sh
+                  # shellcheck disable=SC2046,SC2116
+                  ${FLY_CMD} -t "${FLY_TARGET}" set-team -n "${FLY_TEAM}" \
+                    --github-auth-client-id "${GITHUB_CLIENT_ID}" \
+                    --github-auth-client-secret "${GITHUB_CLIENT_SECRET}"  \
+                    $(echo "$GITHUB_USERS") --basic-auth-username admin --basic-auth-password "${CONCOURSE_ATC_PASSWORD}" --non-interactive
```
alext, line 1273: I don't think we need to set this up for this team (only for the main team). We never log into this team, it only exists to serve as in indicator for which environment you're logging into (to help avoid prod being confused with dev for example)

concourse/pipelines/create-bosh-concourse.yml
LeePorte: Addressed in https://github.com/alphagov/paas-bootstrap/pull/102

concourse/pipelines/create-bosh-concourse.yml
LeePorte: Addressed in https://github.com/alphagov/paas-bootstrap/pull/102

concourse/pipelines/create-bosh-concourse.yml
alext: Actually, after thinking and discussing further, I think it's better that these params remain, and instead we remove the `github-oauth-secrets` resource entirely. This means that the file in the state bucket is purely managed by the `manage-github-secrets.sh` script, and is much more consistent with how the datadog secrets are managed.

PR #102, opened 2017-11-16
Removes unneeded params and team settings

LeePorte:
## What

Fixes the Comments raised in #101 by @alext 

## How to review

1. Create an oAuth application at https://github.com/settings/applications/new
1. Authorization callback URL should be `https://deployer.<your env>.dev.cloudpipeline.digital/auth/github/callback`
1. Store the client ID and Client secret in your personal pass store under the following locations `github.com/concourse/dev/client_id` and `github.com/concourse/dev/client_secret`
1. Checkout this branch and run `make dev upload-github-oauth GITHUB_PASSWORD_STORE_DIR=<your pass dir> DEPLOY_ENV= <your Env>`
1. Push the pipeline with the following command `BRANCH=$(git rev-parse --abbrev-ref HEAD) make dev pipelines CONCOURSE_TYPE=deployer-concourse CONCOURSE_HOSTNAME=deployer BOSH_INSTANCE_PROFILE=bosh-director-cf CONCOURSE_INSTANCE_TYPE=m4.xlarge CONCOURSE_INSTANCE_PROFILE=deployer-concourse ENABLE_COLLECTD_ADDON=true ENABLE_SYSLOG_ADDON=true ENABLE_GITHUB=true`
1. Run the `create-bosh-concourse` pipeline
1. once it has completed, login to team main using the login with Github button.

## Merging

Github API credentials are already in the persistent environment state buckets. Once merge to master the pipeline needs to be run in each environment.

## Who can review

Not @LeePorte

### DIFF
--- a/concourse/pipelines/create-bosh-concourse.yml
+++ b/concourse/pipelines/create-bosh-concourse.yml
@@ -1218,7 +1218,7 @@ jobs:
           params:
             file: updated-concourse-manifest/concourse-manifest.yml
 
-      - task: add-env-specific-team
+      - task: add-github-oauth
         config:
           platform: linux
           image_resource:
@@ -1239,8 +1239,6 @@ jobs:
             FLY_TARGET: ((deploy_env))
             FLY_CMD: ./paas-bootstrap/bin/fly
             ENABLE_GITHUB: ((enable_github))
-            GITHUB_CLIENT_ID: ((github_client_id))
-            GITHUB_CLIENT_SECRET: ((github_client_secret))
           run:
             path: sh
             args:
@@ -1250,8 +1248,6 @@ jobs:
                 export CONCOURSE_URL="https://${CONCOURSE_HOSTNAME}.${SYSTEM_DNS_ZONE_NAME}"
                 export CONCOURSE_ATC_PASSWORD
                 CONCOURSE_ATC_PASSWORD=$(./paas-bootstrap/concourse/scripts/val_from_yaml.rb secrets.concourse_atc_password concourse-secrets/concourse-secrets.yml)
-                export GITHUB_CLIENT_ID
-                export GITHUB_CLIENT_SECRET
 
                 if [ "${ENABLE_GITHUB}" = "true" ] ; then
                   GITHUB_CI_USERS=$(./paas-bootstrap/concourse/scripts/val_from_yaml.rb ci ./paas-bootstrap/concourse/users/users.yml)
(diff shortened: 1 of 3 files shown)


concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -1267,17 +1265,14 @@ jobs:
...
                     $(echo "$GITHUB_USERS") --basic-auth-username admin --basic-auth-password "${CONCOURSE_ATC_PASSWORD}" --non-interactive
 
                   ./paas-bootstrap/concourse/scripts/fly_sync_and_login.sh
-                  # shellcheck disable=SC2046,SC2116
-                  ${FLY_CMD} -t "${FLY_TARGET}" set-team -n main \
-                    --github-auth-client-id "${GITHUB_CLIENT_ID}" \
-                    --github-auth-client-secret "${GITHUB_CLIENT_SECRET}"  \
-                    $(echo "$GITHUB_USERS") --basic-auth-username admin --basic-auth-password "${CONCOURSE_ATC_PASSWORD}" --non-interactive
+                  ${FLY_CMD} -t "${FLY_TARGET}" set-team -n "${FLY_TEAM}" \
+                    --basic-auth-username admin --basic-auth-password "${CONCOURSE_ATC_PASSWORD}" --non-interactive
```
alext: I'm confused by this new version. It seems to duplicate the lines below...

I'm also not clear why we're calling `fly_sync_and_login` so many times.

As I think about this more, I don't think the oauth stuff belongs in this task at all - it's got nothing to do with creating the env-specific team. I think we should create a new task to setup the oauth stuff.

concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -1239,8 +1239,6 @@ jobs:
             FLY_TARGET: ((deploy_env))
             FLY_CMD: ./paas-bootstrap/bin/fly
             ENABLE_GITHUB: ((enable_github))
-            GITHUB_CLIENT_ID: ((github_client_id))
-            GITHUB_CLIENT_SECRET: ((github_client_secret))
```
alext, line 1243: With these gone they can also be removed from the `pipelines.sh` script. This in turn means that we can remove the `retrieve` action from the `manage_github_secrets.sh` script, and make it something that only handles uploading the secrets. This also means that we won't need to upload the secrets for dev deployments unless we've enabled github auth.

concourse/pipelines/create-bosh-concourse.yml
LeePorte: https://github.com/alphagov/paas-bootstrap/blob/b6b56271879211f95886db6938a92759daba3ad7/concourse/pipelines/create-bosh-concourse.yml#L1280-L1282 is the action performed if ENABLE_GITHUB=false

We could break it into a separate task if desired.

concourse/pipelines/create-bosh-concourse.yml
LeePorte: I feel that it is worth keeping the retrieve in `manage_github_secrets.sh` if only for checking when pushing pipelines that the creds have been uploaded.

PR #104, opened 2017-11-23
[#152962527] Upgrade rubocop/govuk-lint

dcarley:
## What

Because we're quite far behind `govuk-lint` and the version of `rubocop`
that it depends on reports as having a vulnerability by GitHub's new
alerting feature:

> We found a potential security vulnerability in one of your dependencies.
> The rubocop dependency defined in Gemfile.lock has a known low severity
> security vulnerability in version range < 0.48.1 and should be updated.

New violations are fixed in separate commits with details in the commit messages.

## How to review

Initially code review. I'd recommend looking at the commits one-by-one.

I've run the `create-bosh-concourse` pipeline from a Deployer Concourse against an existing environment with these changes.

Based on the files that have changed and the type of changes I don't think it's necessary for the reviewer to repeat these steps or to create an environment from scratch. But use your judgement and talk to me if you're unsure.

## Who can review

Anyone.


LeePorte (approved):
Tests pass without issue

PR #112, opened 2017-12-22
[#147833555] Parse YAML lists as comma-delimited strings

henrytk:
## What

This script is duplicated from paas-cf so has to be maintained here
as well. This change was needed to extract a YAML list and pass it
as a comma-delimited string for use as an environment variable.

## How to review

Review in conjunction with https://github.com/alphagov/paas-cf/pull/1151

## Who can review

Anyone but me, @camelpunch, or @46bit


camelpunch:
@alext helpfully suggested that we use the built-in `join` in terraform inside paas-cf. Closing.

PR #113, opened 2018-01-02
[#153366899] Introduce SYSTEM_DNS_ZONE_ID variable

bandesz:
## What

Set SYSTEM_DNS_ZONE_ID in the Makefile for the different envs.

Instead of passing the root DNS zone name and looking up the zone id in the create certificate script just use the values we already have as Terraform variables. To avoid duplicating configuration we move the zone ids to the Makefile and pass them to Terraform and the cert script.

## How to review

1. Bootstrap a new BOSH + Concourse deployment to a new environment and validate in ACM if the certificate was created successfully.
2. Destroy the new deployment

For the persistent environments IMHO it should be enough if we run the create-bosh-concourse pipelines.

## Before merge

Squash the two commits as the first commit is superseeded by the second one. Keep the second commit's commit message only.

## Who can review

Not @bandesz

### DIFF
--- a/concourse/scripts/create_concourse_cert.sh
+++ b/concourse/scripts/create_concourse_cert.sh
@@ -61,7 +61,7 @@ if [ "null" = "${dns_validation_record}" ] || [ "null" = "${dns_validation_value
   exit 1
 fi
 
-zone_id=$(aws route53 list-hosted-zones-by-name --dns-name dev.cloudpipeline.digital --query 'HostedZones[0].Id' --output text | cut -d '/' -f 3)
+zone_id=$(aws route53 list-hosted-zones-by-name --dns-name "${ROOT_SYSTEM_DNS_ZONE_NAME}" --query 'HostedZones[0].Id' --output text | cut -d '/' -f 3)
 
 echo "Upserting DNS validation record: ${dns_validation_record}"
 
--- a/concourse/scripts/create_concourse_cert.sh
+++ b/concourse/scripts/create_concourse_cert.sh
@@ -61,11 +61,9 @@ if [ "null" = "${dns_validation_record}" ] || [ "null" = "${dns_validation_value
   exit 1
 fi
 
-zone_id=$(aws route53 list-hosted-zones-by-name --dns-name "${ROOT_SYSTEM_DNS_ZONE_NAME}" --query 'HostedZones[0].Id' --output text | cut -d '/' -f 3)
-
 echo "Upserting DNS validation record: ${dns_validation_record}"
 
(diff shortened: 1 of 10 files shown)


concourse/scripts/create_concourse_cert.sh
```diff
@@ -61,7 +61,7 @@ if [ "null" = "${dns_validation_record}" ] || [ "null" = "${dns_validation_value
   exit 1
 fi
 
-zone_id=$(aws route53 list-hosted-zones-by-name --dns-name dev.cloudpipeline.digital --query 'HostedZones[0].Id' --output text | cut -d '/' -f 3)
+zone_id=$(aws route53 list-hosted-zones-by-name --dns-name "${ROOT_SYSTEM_DNS_ZONE_NAME}" --query 'HostedZones[0].Id' --output text | cut -d '/' -f 3)
```
dcarley: We already have the zone ID as a Terraform variable, e.g. for dev:

https://github.com/alphagov/paas-bootstrap/blob/190f2bd02783a80b7f35a1e92108eb0f0669451b/terraform/dev.tfvars#L2

Is it possible to use that instead of introducing a new variable?

dcarley (changes requested):
Tested 👌  for dev in it's current form.

Any thoughts on how we can test for the persistent environments? /cc @alext

concourse/scripts/create_concourse_cert.sh
bandesz: Do you suggest we should parse the tfvars file to get this value?

dcarley: I hadn't thought that far ahead 😆 

It might be easier to pull it out of `*.tfvars` and move it into an environment variable in the `Makefile`, then pass it to Terraform with `-var` like we do for `deploy_env`.

bandesz: I added a second commit which solves this. I created a brand new deployment and then destroyed it successfully.

dcarley:
> Any thoughts on how we can test for the persistent environments? /cc @alext

We decided that it would be possible to test staging or production by manually switching back to the wildcard cert and deleting the ACM cert, but that it probably wasn't necessary to test. I've tested creating a new `DEPLOY_ENV` in CI instead.

dcarley (approved):

dcarley:
Oh, shoot. I only just noticed the updated instruction to squash the commits. Sorry!

bandesz:
@dcarley it doesn't matter as the commit history still makes sense.

PR #129, opened 2018-01-31
[#154249348] Use wildcard ACM certificates for public facing domains

bandesz:
## What

Generate wildcard ACM certificates for public facing domains.

We don't want to purchase new wildcard certs every year at a significant cost.
This will also allow us to have valid certs for dev environments.

In this PR we use the new system ACM certificate for Concourse and delete the original Concourse-specific ACM cert.

We also generate the certificate for us-east-1 as There is ]a restriction with ACM certificates when using CloudFront, the certificate must be generated in the eu-east-1 region](https://docs.aws.amazon.com/acm/latest/userguide/acm-services.html)

## How to review

1. Update and run your create-bosh-concourse pipeline from this branch:
   ```BRANCH=acm-for-public-154249348 make dev deployer-concourse pipelines```
1. Check in the ACM console whether
    * *.${DEPLOY_ENV}.dev.cloudpipelineapps.digital was issued
    * *.${DEPLOY_ENV}.dev.cloudpipeline.digital was issued
    * concourse.${DEPLOY_ENV}.dev.cloudpipeline.digital was deleted
1. In Route 53 there should be only one _???.acm-validations.aws record as _???.${DEPLOY_ENV}.dev.cloudpipeline.digital
1. Run the create-bosh-concourse pipeline again to make sure we don't create the certificates multiple times and to make sure the clean up task doesn't fail on the second run
1. Create and destroy a new _build_ Concourse and make sure it doesn't leave any ACM certs behind

## Who can review

Not @keymon or me.

### DIFF
--- a/terraform/concourse/elb.tf
+++ b/terraform/concourse/elb.tf
@@ -1,5 +1,5 @@
-data "aws_acm_certificate" "concourse" {
-  domain   = "${var.concourse_hostname}.${var.system_dns_zone_name}"
+data "aws_acm_certificate" "system" {
+  domain   = "*.${var.system_dns_zone_name}"
   statuses = ["ISSUED"]
 }
 
@@ -22,7 +22,7 @@ resource "aws_elb" "concourse" {
     instance_protocol  = "tcp"
     lb_port            = 443
     lb_protocol        = "ssl"
-    ssl_certificate_id = "${data.aws_acm_certificate.concourse.arn}"
+    ssl_certificate_id = "${data.aws_acm_certificate.system.arn}"
   }
 
   tags {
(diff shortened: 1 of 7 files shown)


terraform/concourse/elb.tf
```diff
@@ -1,5 +1,5 @@
-data "aws_acm_certificate" "concourse" {
-  domain   = "${var.concourse_hostname}.${var.system_dns_zone_name}"
+data "aws_acm_certificate" "system_acm_cert" {
```
alext: I think the `_acm_cert` in the name here is redundant. It's also inconsistent with the naming in paas-cf.

keymon: I will take care of that

keymon: fixed

alext (approved):
These changes look good. I want to hold off merging until https://github.com/alphagov/paas-cf/pull/1202 is also ready for merge.

PR #134, opened 2018-02-06
 [#154437339] Upgrade bosh director and bosh-aws-cpi

keymon:
[#154437339 Upgrade BOSH director and AWS CPI](https://www.pivotaltracker.com/story/show/154437339)

What?
-----

We want to upgrade to the latest version of bosh and bosh-aws-cpi:
 - [v264.7.0](https://github.com/cloudfoundry/bosh/releases/tag/v264.7.0)
 - [v69](https://github.com/cloudfoundry-incubator/bosh-aws-cpi-release/releases/tag/v69)

We are using 261.4 and 62

In this PR we must configure the mutual TLS for the NATs queue in bosh and all the clients: director, healthmanager, and agents:

 - We generate the certificates using the same approach than in paas-cf, instead of using the bosh v2 interpolation. We will migrate to bosh v2 interpolation in the future.
 - We must enable mutual TLS for bosh nats:
    - The certifiates must have CNs like: `*.client.bosh-internal`, `*.director.bosh-internal`, `*.hm.bosh-internal` or `*.nats.bosh-internal`.
    - The nats server cert must have a SAN for the host that the client use
 - We keep enabled the option [`nats.allow_legacy_agents:`](https://github.com/cloudfoundry/bosh/blob/master/jobs/nats/spec#L46) so the already non TLS agents can still work
 - We also generate a SSL cert for the director. Previously it was generated automatically.

Note: We had some trouble with the certificates due wrong values in the CN and SAN, see the commit messages.

How to review
---------

 * Code review
 * Push `make dev deployer-concourse pipelines DEPLOY_ENV=henry BRANCH=upgrade-bosh-and-cpi SELF_UPDATE_PIPELINE=false`
 * run the `create-bosh-concourse` pipeline:
   * It shall redeploy bosh with the new version
   * It shall redeploy concourse without errors
 * Redeploy CF: Shall redeploy without errors.
 * Check that all vms are working in `make bosh-cli`
 * Optionally: use `make ssh_bosh` and `make ssh_concourse` to check that the nats is using SSL
   * in bosh check `/var/vcap/jobs/nats/*`
   * in concourse check `/var/vcap/bosh/*`
   * check the logs.

Before merge
----------------

Update the last commit when the slow slow slow slow automated build in docker hub finishes https://hub.docker.com/r/governmentpaas/bosh-init/builds/b5e6yokk8yvgm5mjfagchid/

Who?
---

Anyone but @keymon or @henrytk


manifests/bosh-manifest/scripts/generate-bosh-certs.sh
```diff
@@ -0,0 +1,87 @@
...
+tar -xvzf "${CA_TARBALL}" -C "${WORKING_DIR}/out"
+
+cd "${WORKING_DIR}"
+for cert_entry in ${CERTS_TO_GENERATE}; do
+  cn=${cert_entry%%,*}
+  domains=${cert_entry#*,}
+
+  if [ -f "${CERTS_DIR}/${cn}.crt" ]; then
+    echo "Certificate ${cn} is already generated."
+    if [ "${ACTION}" = "rotate" ]; then
```
dcarley, line 79: Should we have a rotate job for this pipeline? Do we have any monitoring of internal cert expiry?

dcarley (commented):
I've been seeing these errors in `/var/vcap/sys/log/nats/nats.log` on BOSH:
```
[681] 2018/02/07 11:06:57.701915 [ERR] 127.0.0.1:34138 - cid:189 - An error occurred while peeking connection: read tcp 127.0.0.1:4222->127.0.0.1:34138: read: connection reset by peer
[681] 2018/02/07 11:06:57.701936 [ERR] 127.0.0.1:34138 - cid:189 - An error occurred while detecting if client requested TLS: Expected header size to be 6, but was 0
[681] 2018/02/07 11:08:49.460866 [ERR] 127.0.0.1:34234 - cid:191 - An error occurred while peeking connection: read tcp 127.0.0.1:4222->127.0.0.1:34234: read: connection reset by peer
[681] 2018/02/07 11:08:49.460886 [ERR] 127.0.0.1:34234 - cid:191 - An error occurred while detecting if client requested TLS: Expected header size to be 6, but was 0
[681] 2018/02/07 11:09:10.569811 [ERR] 127.0.0.1:34318 - cid:193 - An error occurred while peeking connection: read tcp 127.0.0.1:4222->127.0.0.1:34318: read: connection reset by peer
[681] 2018/02/07 11:09:10.569835 [ERR] 127.0.0.1:34318 - cid:193 - An error occurred while detecting if client requested TLS: Expected header size to be 6, but was 0
[681] 2018/02/07 11:14:12.173419 [ERR] 127.0.0.1:34504 - cid:195 - An error occurred while peeking connection: read tcp 127.0.0.1:4222->127.0.0.1:34504: read: connection reset by peer
[681] 2018/02/07 11:14:12.173437 [ERR] 127.0.0.1:34504 - cid:195 - An error occurred while detecting if client requested TLS: Expected header size to be 6, but was 0
```

Everything seems to be working - BOSH can deploy and the VMs show as healthy. I haven't had much luck tracking down the "problem" any further. I can't seem to cause the errors by redeploying. Anyone think it's a concern?

keymon:
Those logs annoy me, I do not know where they come from. It looks like something tried to connect using pure SSL, which nats does not support https://github.com/nats-io/gnatsd/issues/291

I would try to investigate

manifests/bosh-manifest/scripts/generate-bosh-certs.sh
keymon: No, I did not add the rotate job. Shall we?

No, we do not have any monitoring for internal certificates, neither here or in paas-cf.

To be honest, I was thinking that maybe we need a step to test all the generated certificates, to verify expire date, CNs, SAN, etc. For instance, in this case, we can check that the CN matches the format required by nats and that the host alias is registered properly. 

On the other hand, we shall ditch all this logic soon as we move to bosh 2 interpolation.

manifests/bosh-manifest/scripts/generate-bosh-certs.sh
```diff
@@ -0,0 +1,87 @@
...
+
+generate_cert() {
+  _cn="${1}"
+  _domains="${2}"
+  _target_dir="${3}"
+
+  certstrap request-cert \
+    --passphrase "" \
+    --common-name "${_cn}" \
+		${domains:+--domain "${_domains}"}
```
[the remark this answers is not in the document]
keymon: I will replace.

manifests/bosh-manifest/scripts/generate-bosh-certs.sh
dcarley: I'm not sure. It seems like we should have both but I don't know whether this story is the right place. @alext do you have any thoughts?

keymon: I added the jobs to rotate those certs, but we still do not have logic to rotate the bosh-CA. We use the bosh-CA as CA for nats clients.

They are completely safe to rotate and it shouldn't meant downtime.

concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -703,18 +750,30 @@ jobs:
...
+                  cn=${file%.crt}
+                  yaml_key_name=$(echo "${cn}" | tr A-Z-. a-z__)
+
+                  echo "Converting ${cn} certs to YAML"
+                  ../paas-bootstrap/concourse/scripts/file_to_yaml.sh secrets "${yaml_key_name}_key" "${cn}.key" > "${cn}_key.yml"
+                  ../paas-bootstrap/concourse/scripts/file_to_yaml.sh secrets "${yaml_key_name}_cert" "${cn}.crt" > "${cn}_cert.yml"
+                done
+
+                # shellcheck disable=SC2103
+                cd .. || true
```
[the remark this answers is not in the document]
keymon:
```
pushd certs
^-- SC2039: In POSIX sh, 'pushd' is undefined.
In - line 14:
popd
^-- SC2039: In POSIX sh, 'popd' is undefined.
```

:) 

I would replace with a subshell as suggested by shell check

manifests/bosh-manifest/scripts/generate-bosh-certs.sh
alext: We already have a story in the icebox to monitor the internal certs (https://www.pivotaltracker.com/story/show/154917736), I think this can be included in that.

It would be good to cover rotating these certs here, but I think the CA cert is out of scope as that affects CF as well as bosh.

dcarley:
> Those logs annoy me, I do not know where they come from. It looks like something tried to connect using pure SSL, which nats does not support nats-io/gnatsd#291
> 
> I would try to investigate

I didn't make any progress with this. Agreed with Alex that it's probably OK. I'll leave a note in the followup story in case disabling non-TLS makes the problem go away.

keymon:
:+1:

PR #138, opened 2018-02-13
[#154959067] Lower `DEPLOY_ENV` maximum length to 8

henrytk:
## What

We now have an ELB unfortunately named `<env>-cf-router-system-domain`
(your're welcome), so, due to the limit of 32 characters in an ELB name,
we have to limit this variable to 8 characters.

## How to review

`DEPLOY_ENV=12345678 make check-env-vars` should exit zero.
`DEPLOY_ENV=123456789 make check-env-vars` should exit non-zero.

## Who can review

Anyone but me.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 help:
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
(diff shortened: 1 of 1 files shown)


samcrang (approved):
```
$ git rev-parse --abbrev-ref HEAD
deploy-env-max-length
$ DEPLOY_ENV=123456789 make check-env-vars
Makefile:13: *** Sorry, DEPLOY_ENV (123456789) has a max length of 8, otherwise derived names will be too long.  Stop.
$ echo $?
2
$ DEPLOY_ENV=12345678 make check-env-vars
make: `check-env-vars' is up to date.
$ echo $?
0
```

PR #140, opened 2018-02-21
[154437854] Upgrade Bosh CLI to v2

bandesz:
## What

We update the Bosh CLI to v2 and we updated all Bosh commands to use the new CLI version.

Changes:
* We duplicated the bosh_CA.crt file as a separate resource so we can use it easily for `bosh login` commands easily
* Don't generate director_uuid into the deployment manifest (the v2 cli doesn't need it)
* Replace the bosh_pre_destroy.rb scripts with checking `bosh deployments`

Additional changes:
* We had to update paas-collectd-boshrelease (https://github.com/alphagov/paas-collectd-boshrelease/pull/4) as it contained a template expression which failed.
* Don't run terraform destroy if all the resources were already destroyed: we got an ACM cert related error when you run the terraform destroy twice (see commit for details)

## How to review

1. Create a new environment from this branch and run the create-bosh-concourse pipeline
    ```BRANCH=bosh_cli_v2_154437854 make dev deployer-concourse bootstrap```
1. Run the create-bosh-concourse pipeline on the newly Concourse too
1. Destroy the environment by running the destroy-bosh-concourse pipeline
1. Update your existing environment from this branch and run the create-bosh-concourse pipeline
    ```BRANCH=bosh_cli_v2_154437854 make dev deployer-concourse pipelines```

## How to release

1. Review and merge https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/118
1. Update the bosh-cli-v2 images tags to the final Docker image tag (fix the `Update bosh cli commands to use v2` and `Update the bosh-cli command to work with Bosh CLI v2` commits)
1. Review and merge https://github.com/alphagov/paas-collectd-boshrelease/pull/4
1. Replace the temporary commit in the PR with the final collectd release
1. Merge this

## Who can review

Not @chrisfarms or @me.


henrytk:
I think any of the 12 commits that have to be changed as a result of review should be marked as `TMP`. I can only see one of them is.

PR #143, opened 2018-02-26
[#154966289] Disable nats.allow_legacy_agents

bandesz:
## What

Disable nats.allow_legacy_agents.

We want all the agents to connect using only mutual TLS.

Note: The NATS user+password is still needed, see [1].

[1] https://github.com/cloudfoundry/bosh-deployment/blob/a4fdaa9d334775359a8f15ef557e0f4099003ccc/bosh.yml

## How to review

1. Run your create-bosh-concourse pipeline from this branch
    ```BRANCH=disable_allow_legacy_agents_154966289 make dev deployer-concourse pipelines```

2. Run the create-cloudfoundry pipeline

3. Check if there are any errors on the BOSH VM in `/var/vcap/sys/log/nats/*.log`.

I created a new Bosh env from scratch and validated that it was created without errors. I don't think it's worth retesting this as it takes a long time.

## Who can review

Not me.

### DIFF
--- a/concourse/resources/bosh-secrets-keep.yml
+++ b/concourse/resources/bosh-secrets-keep.yml
@@ -1,4 +1,4 @@
 ---
 keep:
-  bosh_nats_password: (( grab secrets.bosh_nats_password ))
+  bosh_mbus_bootstrap_password: (( grab secrets.bosh_mbus_bootstrap_password ))
   bosh_postgres_password: (( grab secrets.bosh_postgres_password ))
--- a/manifests/bosh-manifest/bosh-manifest.yml
+++ b/manifests/bosh-manifest/bosh-manifest.yml
@@ -69,8 +69,7 @@ jobs:
         health_monitor:
           certificate: (( grab secrets.default_hm_bosh_internal_cert ))
           private_key: (( grab secrets.default_hm_bosh_internal_key ))
-      # FIXME: To remove as all agents are redeployed with TLS support
-      allow_legacy_agents: true
+      allow_legacy_agents: false
 
     director:
       address: 127.0.0.1
@@ -101,7 +100,7 @@ jobs:
       resurrector_enabled: false
 
     agent:
-      mbus: (( concat "nats://nats:" secrets.bosh_nats_password "@" $BOSH_FQDN ":4222" ))
+      mbus: (( concat "nats://" $BOSH_FQDN ":4222" ))
 
     registry:
       db: (( grab meta.rds ))
--- a/manifests/bosh-manifest/bosh-manifest.yml
+++ b/manifests/bosh-manifest/bosh-manifest.yml
@@ -143,11 +143,11 @@ cloud_provider:
     user: vcap
     private_key: .ssh/bosh_id_rsa
 
-  mbus: (( concat "https://mbus:" secrets.bosh_nats_password "@" $BOSH_FQDN_EXTERNAL ":6868" ))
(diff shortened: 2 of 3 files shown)


dcarley:
> Note: The NATS user+password is still needed, see [1].
> 
> [1] https://github.com/cloudfoundry/bosh-deployment/blob/a4fdaa9d334775359a8f15ef557e0f4099003ccc/bosh.yml

I think that's because they are still using the default of `nats.allow_legacy_agents: true`.

When disabled, [`nats.user` and `nats.password` aren't used](https://github.com/cloudfoundry/bosh/blob/v264.7.0/jobs/nats/templates/nats.cfg.erb#L9-L11) and the credentials can be removed from `agent.mbus` because mTLS is being used instead. The following deployed successfully for me:

```patch
diff --git a/manifests/bosh-manifest/bosh-manifest.yml b/manifests/bosh-manifest/bosh-manifest.yml
index 5e28da2..01e2069 100644
--- a/manifests/bosh-manifest/bosh-manifest.yml
+++ b/manifests/bosh-manifest/bosh-manifest.yml
@@ -53,8 +53,6 @@ jobs:
   properties:
     nats:
       address: 127.0.0.1
-      user: nats
-      password: (( grab secrets.bosh_nats_password ))
       tls:
         ca: (( grab secrets.bosh_ca_cert ))
         client_ca:
@@ -100,7 +98,7 @@ jobs:
       resurrector_enabled: false

     agent:
-      mbus: (( concat "nats://nats:" secrets.bosh_nats_password "@" $BOSH_FQDN ":4222" ))
+      mbus: (( concat "nats://" $BOSH_FQDN ":4222" ))

     registry:
       db: (( grab meta.rds ))
```

Reading [this thread in #bosh](https://cloudfoundry.slack.com/archives/C02HPPYQ2/p1519332872000166?thread_ts=1519316547.000813&cid=C02HPPYQ2) clarified for me that these properties are used separately to create an endpoint that a newly bootstrapped BOSH Director can contact back to `bosh create-env`:

```yaml
cloud_provider:
  mbus: (( concat "https://mbus:" secrets.bosh_nats_password "@" $BOSH_FQDN_EXTERNAL ":6868" ))

  properties:
    agent:
      mbus: (( concat "https://mbus:" secrets.bosh_nats_password "@0.0.0.0:6868" ))
```

I think we should generate a new password called something like `bosh_bootstrap_nats_password` to clarify the difference and [match cf-deployment upstream](https://github.com/cloudfoundry/bosh-deployment/blob/a4fdaa9d334775359a8f15ef557e0f4099003ccc/bosh.yml#L121-L125).

bandesz:
@dcarley nice investigation! I looked through the full Bosh documentation - of course nothing was there about this. Next time I look into the source instead :) I added two extra commits.

Also I used the upstream naming, because `nats` might be confusing in the variable name as there is no NATS when bootstrapping Bosh.

bandesz:
@dcarley The secret variable rename will cause Bosh to be recreated as the `bosh create-env` won't be able to connect to the agent. Is this ok for us?

dcarley:
> @dcarley The secret variable rename will cause Bosh to be recreated as the bosh create-env won't be able to connect to the agent. Is this ok for us?

That seems fine. We can live without a director for a few minutes in return for easier to understand code.

I was going to ask you to rejig the commit messages slightly to explain why the credentials can be removed from `agent.mbus`, but since you're ill I took the liberty of doing it for you. Hope I've done you justice.

PR #156, opened 2018-04-17
[#156228712] Fix YAML structure

samcrang:
We got this a bit wrong and it fails to deploy cleanly in CI. Turns out it's a list of objects with the `user` key not just a list of usernames.

## What

Fix some broken YAML.

## How to review

Eyeball it.

## Who can review

Anyone.


alext:
Doh!  My bad, sorry.

PR #157, opened 2018-04-19
[#154957353] Remove s3init.sh

bandesz:
## What

Remove usage of our ancient `s3init.sh` script in favour of self-initialising S3 resources, which we implement in https://github.com/alphagov/paas-s3-resource/pull/8.

## How to review

Code review should suffice. We have tested migrating, creating and destroying dev environments with this.

🐝 **Merge the s3-resource PR and update this PR's `[TMP]` commit with the DockerHub tag of the resulting container** 🐝

## Who can review

Not @46bit @bandesz.


samcrang:
Updated to the "real" version of `s3-resource`.

samcrang:
Updated temporary commit.

alext:
Do we still need `concourse/init_files/*`, or should they be removed?

46bit:
It looks like at least one use remains https://github.com/alphagov/paas-bootstrap/blob/6bf510b89b5257e4fa4454c66c2c55a31c850952/concourse/pipelines/destroy-bosh-concourse.yml#L438.

PR #159, opened 2018-04-25
[#154957353] Cleanup old S3 init_files

alext:
## What

This removes the last usage of the init file for `bosh create-env` in
the destroy pipeline, and removes all the init files. They're not needed
any more now that we have support for initial versions in the S3
resource.

## How to review

Code review may be enough. You could test deploying and tearing down a bosh/concourse.

## Who can review

Not me.


concourse/pipelines/destroy-bosh-concourse.yml
```diff
@@ -435,7 +435,9 @@ jobs:
                   --state bosh-init-working-dir/bosh-manifest-state.json
                 # If the delete is successful, the file will be missing
                 if [ ! -f bosh-init-working-dir/bosh-manifest-state.json ]; then
-                  cp paas-bootstrap/concourse/init_files/bosh-init-state.json.tpl bosh-init-working-dir/bosh-manifest-state.json
+                  # ID has to match value from the seed in the bosh-cli-v2 container:
+                  # https://github.com/alphagov/paas-docker-cloudfoundry-tools/blob/master/bosh-cli-v2/bosh_init_cache/minimal-state.json
+                  echo '{ "installation_id": "44f01911-a47a-4a24-6ca3-a3109b33f058" }' > bosh-init-working-dir/bosh-manifest-state.json
```
bandesz, line 440: Instead of doing the reset manually we could get the `bosh-init-state` resource with initial version and put it back.

```
- get: bosh-init-state
  version: "-"
- put: bosh-init-state
  params:
    file: bosh-init-state/((bosh_manifest_state))
```

What do you think?

bandesz: After reading the Concourse `get` docs I think my approach might cause an issue on existing dev envs (?):

> If set to a specific version (e.g. {ref: abcdef123}), only that version will be used. Note that the version must be available and detected by the resource, otherwise the input will never be satisfied. You may want to use fly check-resource to force detection of resource versions, if you need to use an older one that was never detected (as all newly configured resources start from the latest version).

alext: Possibly not, but I think it's good for completeness to ensure the stale bosh-init state is replaced. It prevents issues if things error later in the pipeline etc.

Using the initial version would require configuring it in this pipeline as well, so would only move this definition (assuming that approach works as per your comment).

bandesz: 🙌

bandesz (approved):

PR #164, opened 2018-05-23
[#153119909] Upgrade Ruby to 2.5

bandesz:
## What

Update Ruby version to 2.5

We decided to use the same (latest) Ruby version everywhere. We still used 2.2 in multiple places which reached EOL on 2018-03-31 [1], so we decided to migrate to the latest stable version.

We had to upgrade govuk-lint as older versions didn't support Ruby 2.5. Thanks to the upgrade we've got new style check violations but we decided to ignore them (at least for now), as this would unnecessarily blow up the story and increase the risk.

[1] https://www.ruby-lang.org/en/downloads/branches/

## How to review

1. Create a BOSH+Concourse env from this branch:

    ```
    BRANCH=upgrade_ruby_153119909 SELF_UPDATE_PIPELINE=false ENABLE_DATADOG=true make dev deployer-concourse bootstrap
    ```

1. Review https://github.com/alphagov/paas-cf/pull/1366

1. Run the create-bosh-concourse pipeline again on this env or on your existing env.

1. Test `DEPLOY_ENV=... make dev bosh-cli`

1. Test `DEPLOY_ENV=... AWS_ACCOUNT=dev ./scripts/bosh vms`

1. Destroy the env

## How to merge

❗️ The PR contains a WIP commit which needs to be updated with the final Docker image tags.

## Who can review

Not me.


samcrang:
I've updated the commit with real docker images.

PR #166, opened 2018-06-04
[#157047494] Migrate away from runtime config

alext:
## What

We want to move away from using runtime-config to reduce the coupling between this repo and paas-cf.

This removes runtime config, and moves the config that was there into the individual bosh and concourse manifests. There's a companion PR (alphagov/paas-cf#1382) that will add the necessary config to the CF manifest.

Note: instead of migrating the collectd addong, I've chosen to remove it. This is because we're about to play a story to remove Graphite entirely, so there didn't seem to be much point in migrating it.

## How to review

* Deploy from this branch with Datadog enabled.
* Verify you still get metrics for bosh and concourse in Datadog
* Verify you still get logs from Concourse in logsearch

## :rotating_light: Note about merging :rotating_light: 

The merging and deployment of this must be coordinated with the alphagov/paas-cf#1382

## Who can review

Not me.


henrytk:
Based on code review it looks like the syslog-forwarder used to be on the Bosh VM, but now isn't. Am I reading that correctly?

henrytk:
The generate-bosh-config job fails for me with:

```
Error reading file ./paas-bootstrap/manifests/bosh-manifest/addons/datadog-agent.yml: open ./paas-bootstrap/manifests/bosh-manifest/addons/datadog-agent.yml: no such file or directory
```

Just investigating why now...

henrytk:
Not seeing any metrics for my Bosh in Datadog. I can see the collector is logging the following:

```
ConnectionError: ('Connection aborted.', error(97, 'Address family not supported by protocol'))
```

which usually indicates an IP version issue. I can see metrics for your Bosh @alext, so possibly just something weird in my environment.

PR #168, opened 2018-06-19
[#156164640] Create staging and prod London deployments.

alext:
## What

This adds new targets for deploying staging and prod in the London region.

## How to review

Other than code review, the best way to review this is to probably test deploying the London staging deployment. The initial bootstrapping of a persistent env is a little fiddly. Given this it's probably best to pair with @henrytk or myself while deploying.

* Load staging AWS keys and checkout this branch
* Checkout the credstore branch https://github.com/alphagov/paas-credentials/pull/121
* Spin up a bootstrap concourse pointing at this branch: `BRANCH=london make stg-lon deployer-concourse bootstrap`
* Run the `create-bosh-concourse` pipeline (this may fail due to missing secrets)
* Run the secret upload tasks (now that the state bucket exists): `make stg-lon upload-datadog-secrets` and `make stg-lon upload-github-oauth`
* Login to the created deployer concourse, and run the `create-bosh-concourse` pipeline a second time.
* Verify it completes successfully
* Continue with review steps in https://github.com/alphagov/paas-cf/pull/1403

## After merging

* Set the target branch for stg-lon back to master: `make stg-lon deployer-concourse pipelines`
* Deploy the `prod-lon` deployment as above, but don't set the `BRANCH` when creating the bootstrap concourse.

## Who can review

Not @henrytk or myself.

PR #169, opened 2018-06-21
[#156164640] Sign #168

bandesz:
## What

I accidentally merged #168 without signing, so this is an empty commit to add the signature.

[#156164640] Create staging and prod London deployments.


keymon (approved):
Just merge it!

PR #170, opened 2018-06-22
[#157838914] Upgrade Bosh

bandesz:
## What

We have encountered an issue [1] of a process putting a VM under load that causes
monit to drop the connections. This is fixed in the 266.x series, therefore we
should upgrade to that.

Version 266.6.0 may have fixed another issue as well[2].

[1] cloudfoundry/bosh#1754
[2] cloudfoundry/bosh#1990

## How to review

1. Use this branch to update your pipelines:

  ```
  BRANCH=upgrade_bosh_266_157838914 make dev deployer-concourse-pipelines
  ```
1. Run create-bosh-concourse pipeline. It should redeploy Bosh.
1. Run the create-cloudfoundry pipeline to ensure Bosh still does its job.

## Who can review

Not me or @tlwr

### DIFF
--- a/manifests/bosh-manifest/bosh-manifest.yml
+++ b/manifests/bosh-manifest/bosh-manifest.yml
@@ -172,8 +172,8 @@ resource_pools:
 - name: bosh
   network: private
   stemcell:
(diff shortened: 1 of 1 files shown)


alext:
I'm going to mark this as WIP while we work out the issues with links etc in paas-cf.

PR #172, opened 2018-07-09
Cleanup task to delete old ACM certs

alext:
## What

This was left over from when we switched the concourse machines to use
the wildcard system domain cert. These old certs have all been cleaned
up, so this can be removed.

I've checked in the ACM console, and we no longer have any certs left to cleanup.

## How to review

Code review is probably enough.

## Who can review

Not me.


keymon:
~~Shouldn't we also delete `./paas-bootstrap/concourse/scripts/delete_acm_cert.sh`?~~

the delete script is also used in the destroy pipeline

keymon (approved):
LGTM

PR #176, opened 2018-07-18
[#157769641] Remove syslog-forwarder in concourse

keymon:
https://www.pivotaltracker.com/n/projects/1275640/stories/157769641

What
----

Currently we are forwarding all the logs to logit from both the concourse and the CF deployments (although disabled for the concourse build deployment).

But as we share logit stacks for each environment (dev, staging, ci and prod), all the logs for different deployments end in the same shared Elasticsearch index.

In the case of CF we incidentally called our deployment after the $DEPLOY_ENV variable (e.g. `hector`, `henry`, `staging`, `prod-lon`) and we can filter the logs using the `@source.deployment` field in kibana.

But in the case of concourse, we call the deployment `concourse` and there is not a practical way to differenciate the logs.

Ideally we would have an additional dedicated tag for all the logs, likely based on $DEPLOY_ENV, to differenciate deployments. But the reality is that we do not use the concourse logs in logit, it is a quite simple deployment with only one VM.

Because that we have decided to simply stop forwarding to logit all the logs from concourse, and we will implement the additional tag feature later in the future if needed.

In this commit we disable this by removing all the configuration to add the syslog-forwarder in concourse, as we do not want to keep code that is not being used.

> Note: In order to force a redeployment of the concourse VM,so all the rsyslog files copied in `/etc/rsyslog.d` are deleted, we do a minor bump of the stemcell. That would force bosh to recreate the VM.

How to review
-------------

 - Code review
 - Deploy this and check that no more logs are being sent to logit  anymore: 
    1. push the pipeline `make dev deployer-concourse pipelines DEPLOY_ENV=...  BRANCH=$(git rev-parse --abbrev-ref HEAD) `
    2. Rerun the pipeline. It takes some minutes, go for a coffee. The concourse job would error because it kill concourse itself :)
    3. Rerun the concourse job.
    4. Go to logic and search for `@source.id:<vmid>`.  The vm id is in the concourse deploy task.

Follow up
-------------

After merge, apply this to staging and prod.

Who can review
--------------

Anyone but me


alext:
Looks good to me.

With this change, I wonder whether the `upload-logit-secrets` make task should move to paas-cf. What do you think?

That can be a separate PR though, so I'm going to merge this.

keymon:
> With this change, I wonder whether the upload-logit-secrets make task should move to paas-cf. What do you think?

We have all the `upload-*` here, so I guess it is fine. 

You know me: We need to find a better way of handle these kind of data, instead generate .yml files with bash, upload them to s3, and them parse them adhoc. :smile:  . Otherwise, I would keep all together.

PR #182, opened 2018-09-10
[#159730005] Enable bosh-dns in the director's runtime-config

jpluscplusm:
What
----

This PR, which must be closely coordinated with (and **must** come before) the merge and deployment of alphagov/paas-cf#1522, makes the deployment of bosh-dns global across all deployments from each environment's BOSH director.  This is a prerequisite which upstream (cf-deployment 3.0.0+) insist upon.

For operational reasons, this PR should be merged and deployed *after* alphagov/paas-cf#1523.

Our process is slightly complicated by the fact that we have to pull the existing bosh-dns CA and non-CA certs for healthchecks and the bosh-dns API over from CF, where it's been deployed for a few weeks, and adopt it in the `create-bosh-concourse` pipeline. This is because we're unsure if the API is exposed between VMs, or just inside each VM, hence if having half the system using one CA and half using another would be problematic. We *are* using our standard pattern to enable CA rotation, but it doesn't feel like the right place to prove its correctness, along with the CA migration across pipelines/systems.

Review
-------

This has been paired on throughout, and is being rolled out by @tlwr, @richardTowers and myself, with @henrytk  helping from a production access perspective.

Deployment plan: https://hackmd.cloudapps.digital/nZAnzvldRLqtxcEXbXoa6w?view#Deloyment-to-Production

### DIFF
--- a/concourse/pipelines/create-bosh-concourse.yml
+++ b/concourse/pipelines/create-bosh-concourse.yml
@@ -261,6 +261,21 @@ resources:
       versioned_file: concourse_id_rsa
       initial_version: "-"
 
+  - name: cf-vars-store
+    type: s3-iam
+    source:
+      bucket: ((state_bucket))
+      region_name: ((aws_region))
+      versioned_file: cf-vars-store.yml
+
+  - name: bosh-runtime-config-vars
+    type: s3-iam
+    source:
+      bucket: ((state_bucket))
+      region_name: ((aws_region))
+      versioned_file: bosh-runtime-config-vars.yml
+      initial_version: "-"
+
 jobs:
   - name: init-bucket
     serial: true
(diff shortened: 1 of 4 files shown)


tlwr:
As above, as of https://github.com/alphagov/paas-cf/pull/1523 this is ZDD in combination with https://github.com/alphagov/paas-cf/pull/1522 and is ready to go

concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -261,6 +261,21 @@ resources:
       versioned_file: concourse_id_rsa
       initial_version: "-"
 
+  - name: cf-vars-store
+    type: s3-iam
+    source:
+      bucket: ((state_bucket))
+      region_name: ((aws_region))
+      versioned_file: cf-vars-store.yml
```
henrytk, line 269: This name is now confusing, because it is a variables store used by Bosh, and therefore may have nothing to do with Cloud Foundry (`cf-`) (we use this pipeline to deploy our CI environment as well as Cloud Foundry).

I am aware this has already been merged ;)

alext: This actually does have to do with CF. It's pulling in the CF vars store so that we can extract the existing bosh-DNS certs from it as part of migrating bosh-DNS here.  Once the migration has settled, there will be a follow-up to remove the usage of this.

PR #184, opened 2018-09-12
[#159730005] Generate bosh runtime-config variables in the bosh update-runtime-config task

tlwr:
tldr;

- do not copy cf-vars-store anymore
- do not have a separate task for bosh-runtime-config interpolation, do it in bosh update-runtime config step instead.

---

# What

Previously we had a task `steal-cf-vars` which copied bosh-dns related
pki material from `cf-vars-store` to `bosh-runtime-config-vars-store`

Now that bosh-dns has been rolled out everywhere (that we care about:
staging, prod) we can get rid of the copying task. This task is now
inlined into the runtime config task, which generates said variables if they do not exist.

Expected behaviour is that new environments have the vars generated, and
existing environments use the existing vars.

# How to review

- Code review

#### First

- Take a copy of `bosh-runtime-config-vars-store.yml`
- Run `create-bosh-concourse`
- Ensure that the copy of `bosh-runtime-config-vars-store.yml` is the same

#### Then

- Delete `bosh-runtime-config-vars-store.yml`
- Run `create-bosh-concourse`
- Green


concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -700,6 +694,38 @@ jobs:
...
+            args:
+              - -ec
+              - |
+                bosh interpolate \
+                  --var-errs \
+                  --vars-store bosh-runtime-config-vars-store/bosh-runtime-config-vars-store.yml \
+                  paas-bootstrap/manifests/runtime-config/variables-template.yml
+
+                cp bosh-runtime-config-vars-store/bosh-runtime-config-vars-store.yml \
+                   new-bosh-runtime-config-vars-store/bosh-runtime-config-vars-store.yml
```
alext: Is there a reason this needs to be a separate task like this. Could it not be part of the set-runtime-configs task? It would make sense to me to have all the needed variables defined in the runtime-config manifest, and get rid of the `variables-template.yml` file. This would be more consistent with other manifests (eg the CF manifest).

tlwr:
Addressed the above comments, updated the PR description.

jpluscplusm (approved):
LGTM.

~~Suggestions:~~

~~- include `BOSH_NON_INTERACTIVE` in the remaining task's envvars.~~

richardTowers (approved):

concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -1042,10 +1007,17 @@ jobs:
                   --name '' \
                   blank-runtime-config.yml
 
+                cp bosh-runtime-config-vars-store/bosh-runtime-config-vars-store.yml \
+                   new-bosh-runtime-config-vars-store/bosh-runtime-config-vars-store.yml
+
                 bosh update-runtime-config \
```
keymon: I thik it is good to still do two steps calling `bosh interpolate` and then `bosh update-runtime-config`.

The reason is that in `bosh interpolate` you can pass this two handy parameters:

```
          --var-errs             Expect all variables to be found, otherwise error
          --var-errs-unused      Expect all variables to be used, otherwise error
```

manifests/runtime-config/bosh-dns.yml
```diff
@@ -35,3 +35,62 @@ addons:
...
+
+- name: dns_api_client_tls
+  type: certificate
+  options:
+    ca: dns_api_tls_ca
+    common_name: api.bosh-dns
+    extended_key_usage:
+    - client_auth
+
+## Our additions
```
keymon: I wouldn't put that comment. It would get forgotten :)

manifests/runtime-config/bosh-dns.yml
```diff
@@ -35,3 +35,62 @@ addons:
...
+
+- name: dns_healthcheck_tls_ca_old
```
keymon: BTW, I am not able to find where we implement the certificate rotation for this project. We do in paas-cf, but not sure if we do here. 

Should we include these variables if they are not useful?

concourse/pipelines/create-bosh-concourse.yml
jpluscplusm: Thanks, @keymon - I *knew* that `update-runtime-config` doesn't accept `--vars-errs` and that was why I'd broken it into 2 steps (or, at least, 2 commands), but I didn't flag it up in review yesterday.  Good call.

manifests/runtime-config/bosh-dns.yml
jpluscplusm: These vars are needed to satisfy the static analysis tests in Travis, which insist on all `*ca*` things having an `_old` variant concatenated.  We could get by with dummy values in there, not backed by bosh-generated "proper" CAs, but it seemed like more work to do that than simply to include them as carbon copies of their non-`_old` siblings.

keymon: Oh, do we have such test in this project?

concourse/pipelines/create-bosh-concourse.yml
jpluscplusm: @keymon Addressed in the recent push.

manifests/runtime-config/bosh-dns.yml
jpluscplusm: I really thought we did, but it's very possible we'd assumed such a test existed, based on paas-cf.

Given the `_old` vars are in production right now, I don't think the overhead of having them there is significant - it would be more effort to remove them.

I don't feel this particularly /strongly/, however :-) @alext @richardTowers @henrytk opinions?

keymon: IMO I would delete the `_old` if it is not being tested and the feature to rotate does not exist. It is misleading and it does not provide anything.

Also it is easy to remove and to add later if needed.

richardTowers: I think let's remove them.

keymon (approved):
Despite the `_old` cert, this looks good for me.

jpluscplusm:
Addressed all comments.

manifests/runtime-config/bosh-dns.yml
jpluscplusm: @richardTowers @keymon Done. Please re-review.

keymon: LGTM. I am rerunning in my env and would merge soon.

keymon:
> You can retrieve the config from the bosh director with bosh config runtime --name=bosh-dns

Yes, as you say that is the **current** config, but we do not store a history of the configurations. 

I think that now we are adding the bosh-dns here, this is a critical piece of config and we should be able to audit it if needed. We should know why a changed was triggered and what changed.

jpluscplusm:
@keymon 
> history of the configurations

That's "this repo", isn't it? Apart from pushing a version containing the certs, which feels a bit odd from a secrets perspective, what would storing the deployed manifest add to our view of history?

PR #189, opened 2018-09-20
[#160614327] Update bosh-cli-v2 image to use v5.2.1

keymon:
What
----

We want to use the bosh-cli v5.2.1 that includes support to
generate CA certificates with SKI[1], and stop using our
fork.

This new container includes the latest PR [2][3]

[1] https://github.com/cloudfoundry/bosh-cli/releases/tag/v5.2.1
[2] https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/134
[3] https://hub.docker.com/r/governmentpaas/bosh-cli-v2/builds/bdqhaxoavpkqqlgq2ivyg5k/

Dependencies
--------------

See https://github.com/alphagov/paas-cf/pull/1552

How to review
-------------

Code review. This has been tested in [2] and in the pipeline of alext.

Who can review
--------------

not me

### DIFF
--- a/concourse/pipelines/create-bosh-concourse.yml
+++ b/concourse/pipelines/create-bosh-concourse.yml
@@ -947,7 +947,7 @@ jobs:
             type: docker-image
             source:
               repository: governmentpaas/bosh-cli-v2
-              tag: 0eff5b6a9c092f865a2b19cc4e75a3b539b82fa2
+              tag: 4012d25ceb903b46908a830b8e05773ced1c8f86
           inputs:
             - name: bosh-manifest
             - name: bosh-init-state
--- a/concourse/pipelines/destroy-bosh-concourse.yml
+++ b/concourse/pipelines/destroy-bosh-concourse.yml
@@ -220,11 +220,11 @@ jobs:
       - task: destroy-concourse
         config:
           platform: linux
-          image_resource:
+          image_resource: &gov-paas-bosh-cli-v2-image-resource
             type: docker-image
             source:
               repository: governmentpaas/bosh-cli-v2
-              tag: 0eff5b6a9c092f865a2b19cc4e75a3b539b82fa2
+              tag: 4012d25ceb903b46908a830b8e05773ced1c8f86
           inputs:
           - name: paas-bootstrap
           - name: bosh-secrets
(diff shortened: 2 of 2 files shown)


concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -343,7 +343,7 @@ jobs:
             type: docker-image
             source:
               repository: governmentpaas/self-update-pipelines
-              tag: 0eff5b6a9c092f865a2b19cc4e75a3b539b82fa2
+              tag: 4012d25ceb903b46908a830b8e05773ced1c8f86
```
alext: I don't think you meant to update this container - self-update-pipelines won't have changed, and this will only cause an unnecessary download in Concourse.

scripts/bosh
```diff
@@ -54,4 +54,4 @@ docker run  \
 	--env "BOSH_ENVIRONMENT=bosh.${SYSTEM_DNS_ZONE_NAME}" \
 	--env "BOSH_CA_CERT" \
 	--env "BOSH_DEPLOYMENT=${DEPLOY_ENV}" \
-	governmentpaas/bosh:0eff5b6a9c092f865a2b19cc4e75a3b539b82fa2 "$@"
+	governmentpaas/bosh:4012d25ceb903b46908a830b8e05773ced1c8f86 "$@"
```
alext: This container hasn't been updated, and is still using 3.0.1. Maybe we should update it as well so that we're using a consistent version everywhere.

scripts/bosh-cli.sh
```diff
@@ -26,4 +26,4 @@ docker run \
     --env "BOSH_ENVIRONMENT=bosh.${SYSTEM_DNS_ZONE_NAME}" \
     --env "BOSH_CA_CERT" \
     --env "BOSH_DEPLOYMENT=${DEPLOY_ENV}" \
-    governmentpaas/bosh-shell:0eff5b6a9c092f865a2b19cc4e75a3b539b82fa2
+    governmentpaas/bosh-shell:4012d25ceb903b46908a830b8e05773ced1c8f86
```
alext: Ditto here, this is still using an old version (2.0.48)...

scripts/bosh
alext: As an aside, I can't figure out where we use this script/container instead of the bosh-shell one below. It feels as though this is redundant, and could probably be removed.

alext (changes requested):
See comments below.

scripts/bosh
keymon: That was added by @henrytk in c1bd7cfd8cdfda8a09af117d908a39a02e7910ee, better ask him

scripts/bosh-cli.sh
keymon: See: https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/135

PR #192, opened 2018-09-21
[#158180318] Upgrade Concourse to 4.2.1

jpluscplusm:
What
----

This PR upgrades the version of Concourse that our bootstrap, deployer and builder CI instances use to version 4.2.1.

Some yaks were shaved during this: 

- upgrade the bootstrap Ubuntu (to Bionic) to fetch a recent docker-compose version
- fix a race condition with cloud-init updating apt's `sources.list`
- modify the fly login wrapper to cope with the Concourse v4 auth changes
- modify the bootstrap pipeline to consume an upstream repo which specifies a semi-official concourse/garden-runc/postgres version tuple

Despite Concourse's major version bump, which was necessary due to the auth changes, the underlying CI hasn't changed /too/ much in v4. A few UI tweaks have happened - the most notable being a dashboard on the `/` of a Concourse install instead of one of the pipelines.

How to review
-------------

- Code review
- Bootstrap: stand up a new bootstrap CI from this branch, and use it to stand up a deployer CI and a CF. Export `BRANCH` so that your bootstrap CI looks at this branch for the env it creates.
- Release: already proven on the `alex-bld` builder CI. 
  - merge after https://github.com/alphagov/paas-release-ci/pull/79
- Deployer: already proven on the `jonmtest` deployer CI + slim dev CF env.
  - merge after https://github.com/alphagov/paas-cf/pull/1555

Who can review
--------------

@jpluscplusm and @leeport paired on most of this; @jpluscplusm solo'd on the tail end. J+L are going to pull in code review from @alext and then review, merge and deploy it as a pair.


manifests/concourse-manifest/spec/manifest_generation_spec.rb
```diff
@@ -35,7 +35,7 @@ def merge_fixtures(fixtures)
 
   it "gets values from secrets" do
     expect(
-      atc_job.fetch("properties").fetch("basic_auth_password")
+      atc_job.fetch("properties").fetch("add_local_users")[0].sub(/^[^:]+:/, '')
```
alext: I wonder whether this would be clearer using split - would save having to mentally parse a regexp:
```ruby
atc_job.fetch("properties").fetch("add_local_users")[0].split(':', 2)[1]
```

concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -269,6 +269,12 @@ resources:
       versioned_file: bosh-runtime-config-vars-store.yml
       initial_version: "-"
 
+  - name: concourse-bosh-deployment
+    type: git
+    source:
+      uri: https://github.com/concourse/concourse-bosh-deployment
+      tag_filter: "v4.2.*"
```
alext: Can we pin the exact version of concourse we're using. We currently rely on the property that deploying a specific revision of paas-bootstrap will result in a consistently configured setup whenever it's run.

concourse/scripts/environment.sh
```diff
@@ -65,4 +65,9 @@ export BOSH_FQDN_EXTERNAL=${BOSH_FQDN_EXTERNAL}
 export VAGRANT_SSH_KEY=${PROJECT_DIR}/${DEPLOY_ENV}-vagrant-bootstrap-concourse
 export FLY_CMD=${FLY_CMD}
 export FLY_TARGET=${FLY_TARGET}
+export CONCOURSE_POSTGRES_DATABASE=${CONCOURSE_DATABASE_NAME}
+export CONCOURSE_POSTGRES_USER=${CONCOURSE_DATABASE_USER}
+export CONCOURSE_POSTGRES_PASSWORD=${CONCOURSE_DATABASE_PASS}
+export CONCOURSE_ADD_LOCAL_USER=${CONCOURSE_ATC_USER}:${CONCOURSE_ATC_PASSWORD}
+export CONCOURSE_EXTERNAL_URL=${CONCOURSE_URL}
```
alext: I'm not clear why these variables need to be added to the environment script given they have the same values as the similarly named variables already present here. I fell this duplication has the potential to lead to confusion.

alext (changes requested):
Generally looks good. I have a few comments; details below:

concourse/scripts/environment.sh
jpluscplusm: This is to allow us to upstream-PR the docker-file changes that consume them, in a way which upstream is hopefully likely to find acceptable.

manifests/concourse-manifest/spec/manifest_generation_spec.rb
jpluscplusm: Passwords /might/ contain colons. Not with our current secret generator, but assuming that that code is driving this value seems like unnecessary coupling.

I feel like, for safety, we'd need to pop the first element after the split, then join the remaining fields with `:`s. Is that simpler, versus using the current regex and perhaps adding a comment?

alext: I'd accounted for that - the `2` param to `split` constrains it to at most 2 results:
```ruby
"foo:bar:baz".split(':',2)[1] # => "bar:baz"
```

alext (approved):

alext:
Looks good to me from a code-review PoV. I'll leave it to you to merge and deploy when you're ready.

PR #193, opened 2018-09-26
[#160727627] Remove temporary concourse team cleanup task

alext:
## What

This was required prior to the upgrade, but is no longer required now
that everything has been upgraded.

How to review
-------------

Run the pipeline, verify it applies cleanly - it should be a no-op deploy.

Who can review
--------------

Not me.

### DIFF
--- a/concourse/pipelines/create-bosh-concourse.yml
+++ b/concourse/pipelines/create-bosh-concourse.yml
@@ -1341,7 +1341,6 @@ jobs:
         - get: bosh-secrets
         - get: concourse-manifest
           passed: ['generate-concourse-config']
-        - get: concourse-secrets
         - get: bosh-CA-crt
           passed: ['generate-secrets']
         - get: concourse-bosh-deployment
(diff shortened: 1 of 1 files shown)


concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -1341,7 +1341,6 @@ jobs:
         - get: bosh-secrets
         - get: concourse-manifest
           passed: ['generate-concourse-config']
-        - get: concourse-secrets
```
jpluscplusm, line 1344: Good catch.

jpluscplusm (approved):
LGTM.

concourse/pipelines/create-bosh-concourse.yml
alext: I didn't catch it, our concourse linter did...

richardTowers (approved):
Looks good to me too. Deployed to my pipeline and it looks like a noop as predicted.

PR #195, opened 2018-10-01
[Firebreak] Use public keys for commit signature verification

henrytk:
## What

We want to avoid reliance on external key servers because they
occassionally fail, causing us to have to spend time investigating it or
retriggering jobs.

We keep using the `.gpg-id` file because accessing the public key ID is
the simplest way for an operator to reference an individual's key. A
simple helper script can be used to generate the lengthy vars file
required by Concourse. Example usage:

```bash
cd paas-cf
ruby concourse/scripts/generate-public-key-vars.rb
```

The vars file is made available unconditionally to all pipelines, but is
gracefully ignored by those who don't use it.

How to review
-------------

Review in conjunction with https://github.com/alphagov/paas-cf/pull/1564

Once done, apply to all environments.

Who can review
--------------

Anyone.


alext (changes requested):
See comments on https://github.com/alphagov/paas-cf/pull/1564

alext:
I've rebased and regenerated the keys file after #197 was merged.

alext (approved):

PR #199, opened 2018-10-15
Rotate Miki's GPG key for a new Yubikeyed one

46bit:
What
----

See commit message.

How to review
-------------

1. Verify this is changing my GPG key from `CAAD002F768C0A74` to `BB97A18439B31647`;
2. Let me know if you know of anywhere else in `paas-bootstrap` that needs changing.

Who can review
--------------

Not @46bit. Ideally someone with context on https://github.com/alphagov/paas-cf/pull/1564 in case I've missed a new Make target in `paas-bootstrap`.


46bit:
This won't work because Pay split up the private keys into separate subkeys. I'll figure out something else.

46bit:
Actually, I think we're golden. But someone will need to update my key ID and resign `paas-credentials`--I can't easily because I've already revoked my old key.

PR #201, opened 2018-10-17
[#160508058] Cleanup after removal of Bosh DNS

henrytk:
What
----

Now we have migrated to deploying bosh-dns from manifest addons we no
longer need this code.

How to review
-------------

Apply on top of https://github.com/alphagov/paas-bootstrap/pull/200 then run the create-cloudfoundry pipeline.

Who can review
--------------

@LeePorte


keymon (approved):
Merging. I think all the environments have been updated.

Otherwise, you can always do:
```
make dev bosh-cli DEPLOY_ENV=...

bosh delete-config -n --type runtime --name bosh-dns
```

PR #210, opened 2018-11-09
[#161544435] Upgrade releases: bosh, cpi, uaa, bpm... and stemcells

keymon:
What
----

We want to upgrade to the latest release of all the components
for bosh.

We also want to upgrade the stemcells to keep them at the same version than cf-deployment v6.0.0. This introduces the Xenial stemcells.

We update stemcells for bosh and concourse.

How to review
-------------

 - Deploy bosh
 - check deployments for concourse and CF still work

Who can review
--------------

Not me


richardTowers:
Code looks fine, running this down my pipeline now.

PR #215, opened 2018-11-17
[#SUPPORT] Deploy patched version of bosh v268.2.0

keymon:
What
----

Bosh v268.2.0 seems to have a bug that occassionally causes bosh
dns to wipe DNS entries during a deployment.

This caused API availability issues in our staging deployment.

Reported upstream in slack[1], they claim that there is a bug related
to a patch[2].

Here we use a patched version of bosh including that patch.

[1] https://cloudfoundry.slack.com/archives/C02HPPYQ2/p1542383747442400
[2] https://github.com/cloudfoundry/bosh/commit/bc4bc20fcf8fec20c975d8d6b1bf0c2

How to review
-------------

Code review

Note: A PR for release-ci is ready in case we want to generate the
release from concourse build, and not use the one generated manually:

https://github.com/alphagov/paas-release-ci/pull/82

If merged, update this PR

Who can review
--------------

Not me

### DIFF
--- a/manifests/bosh-manifest/bosh-manifest.yml
+++ b/manifests/bosh-manifest/bosh-manifest.yml
@@ -20,9 +20,10 @@ meta:
 
 releases:
 - name: "bosh"
-  version: "268.2.0"
-  url: "https://bosh.io/d/github.com/cloudfoundry/bosh?v=268.2.0"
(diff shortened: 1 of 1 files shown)


alext:
I've updated this to use the version built by release-ci (https://github.com/alphagov/paas-release-ci/pull/82)

PR #217, opened 2018-11-21
[#162118725] Upgrade bosh and concourse stemcells to 170.9

alext:
What
----

To fix a bunch of CVEs

How to review
-------------

* Code review is probably enough - I've already run this in my dev environment.

Who can review
--------------

Not me.


henrytk:
What CVEs does this fix? I want to go have a look. Possibly useful to put a link in the commit message.

PR #220, opened 2018-11-27
[#161184357] Upgrade AWS Terraform provider

alext:
What
----

This upgrades the AWS Terraform provider and fixes a workaround that's no longer needed.

This is a precursor to switching to using Terraform to generate ACM certs. The version of the provider we're currently running doesn't support the necessary resources to enable this.

I've also taken the opportunity to DRY the pipeline image_resource definitions.

How to review
-------------

Run the pipeline from this branch, verify everything works.

## 🚨 Before merging 🚨 

* Merge https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/145
* Update the TMP commit.

Who can review
--------------

Not me.

### DIFF
--- a/concourse/pipelines/create-bosh-concourse.yml
+++ b/concourse/pipelines/create-bosh-concourse.yml
@@ -1088,23 +1088,6 @@ jobs:
           params:
             file: updated-concourse-tfstate/concourse.tfstate
 
-      # Temporary task to add the git-${DEPLOY_ENV} user to git group
-      - task: add-git-user-to-group
-        config:
-          platform: linux
-          image_resource: *awscli-image-resource
-          inputs:
-          - name: paas-bootstrap
-          params:
-            DEPLOY_ENV: ((deploy_env))
-          run:
-            path: sh
-            args:
-            - -e
-            - -c
-            - |
-              aws iam add-user-to-group --user-name "git-${DEPLOY_ENV}" --group-name concourse-pool-git-rw
-
   - name: generate-concourse-config
     serial: true
     plan:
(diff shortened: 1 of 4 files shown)


concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -1,4 +1,47 @@
...
+    bosh-cli-v2: &gov-paas-bosh-cli-v2-image-resource
+      type: docker-image
+      source:
+        repository: governmentpaas/bosh-cli-v2
+        tag: 93e168b4244873902cd552ae08fc4a67a7b2e6f5
+    certstrap: &certstrap-image-resource
+      type: docker-image
+      source:
+        repository: governmentpaas/certstrap
+        tag: 895cf6752c8ec64af05a3a735186b90acd3db65a
```
bandesz, line 18: We had two different tags previously and 465642da06051a55630d39c899697b678f66a7f7 is the newer one, so I think we should use that.

bandesz (changes requested):
We should amend the tag for the certstrap image

bandesz:
There is one image reference in concourse-lite-self-terminate.yml. I think we should extract the image resource the same way, so it's consistent in all pipelines.

concourse/pipelines/create-bosh-concourse.yml
alext: I picked the one that was used in more places. If you look at [the diff](https://github.com/alphagov/paas-docker-cloudfoundry-tools/compare/895cf67...465642d), you can see that there are actually no changes to the certstrap container between those versions.

alext:
I've updated the remaining image_resource references...

bandesz (approved):

bandesz:
I amended the tmp commit with the latest terraform image tag.

PR #222, opened 2018-11-28
[#161184357]  Use Terraform to create ACM certs.

alext:
What
----

This moves the ACM cert management into Terraform because it's much more robust than our set of scripts.

As part of this change, we no longer generate the apps cert in this
repo. It's only needed by the CF deployment, so doesn't make sense to
include here - we'll manage it with Terraform in the paas-cf repo. It
was originally added here to avoid having to duplicate the scripts into
paas-cf.

This additionally includes a temporary commit to import the existing cert into terraform to prevent it attempting to create a second one.

How to review
-------------

* Run the pipeline from this branch
* Verify terraform updates correctly and doesn't create a second Cert.
* Run the pipeline a second time to verify the import process is idempotent.

Who can review
--------------

Not me.


bandesz (approved):

concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -1042,6 +1021,34 @@ jobs:
...
+          run:
+            path: sh
+            args:
+            - -e
+            - -c
+            - |
+              tf_cert_details=$(jq -r '.modules[].resources."aws_acm_certificate.system"' concourse-tfstate/concourse.tfstate)
+              if [ "${tf_cert_details}" = "null" ]; then
+                echo "Finding cert to import"
+                aws acm list-certificates \
```
bandesz: Creating a new environment failed, because the `aws acm list-certificates` returns an empty string if the certificate doesn't exist. We have to handle this case too.

alext: Added a check for non-existing cert.

bandesz:
The PR looks good, I tested the following:
 * update an existing env 2x
 * create a new env, update it, then destroy

PR #230, opened 2018-12-27
[#161591823] Use bosh-deployment for bosh manifest

keymon:
What
----

We want to migrate paas-bootstrap bosh-manifest to use bosh-deployment in a similar way we did with cf-deployment in paas-cf.

https://github.com/cloudfoundry/bosh-deployment

In this PR we implement a model similar to the paas-cf prometheus manifest, with a submodule for bosh-deployment.

We work based on https://github.com/alphagov/paas-bootstrap/pull/231 to achieve a nop manifest. 

We move the secret and cert generation into `bosh interpolate`, so that the secrets and certs would be stored in the `bosh-vars-store.yml`. 

We adapt the rotation of secrets, but do not fix the rotation of the bosh-CA. Bosh CA would still be generated using certstrap.

In this PR we do not want to upgrade any software, so we pin the versions to older versions, only for the sake of migrating the pipeline.

See commits for details.

Dependencies
-------------------

Review with https://github.com/alphagov/paas-bootstrap/pull/231 

If https://github.com/alphagov/paas-bootstrap/pull/231 is merged, then rebase this PR.

After: https://github.com/alphagov/paas-bootstrap/pull/232 and https://github.com/alphagov/paas-cf/pull/1695

How to review
-------------

- Review with https://github.com/alphagov/paas-bootstrap/pull/231 
- Code review.
- Check the script `./diff-manifest.sh init` and `./diff-manifest.sh diff` to check that  the manifest is equivalent. The file itself would be different, so the bosh-deploy would redeploy the VM regardless.
- Check the secret rotation: For that you need to revert the migrate commit!
- Remove any WIP commit.

This script https://gist.github.com/keymon/6e38941995a48af109e28192bff6c9cb might be handy to compare versions of files in S3. 
 
Who can review
--------------

Not me.


LeePorte (approved):

manifests/bosh-manifest/scripts/generate-bosh-certs.sh
```diff
@@ -1,89 +0,0 @@
...
-#
-# <name_cert>,<domain1>[,domain2,domain3,...]
-#
-# Note: ALWAYS add a comma after <name_cert>, even if there are no domains
-#
-CERTS_TO_GENERATE="
-default.nats.bosh-internal,bosh.${SYSTEM_DNS_ZONE_NAME}
-default.director.bosh-internal,
-default.hm.bosh-internal,
-bosh_director,bosh.${SYSTEM_DNS_ZONE_NAME},bosh-external.${SYSTEM_DNS_ZONE_NAME}
```
alext, line 19: These SAN's aren't reflected in the cert generated in the manifest. This means that for a clean deployment things fail with cert errors eg:
```
  Performing request GET 'https://bosh.alext.dev.cloudpipeline.digital:25555/info':
    Performing GET request:
      Retry: Get https://bosh.alext.dev.cloudpipeline.digital:25555/info: x509: certificate is not valid for any names, but wanted to match bosh.alext.dev.cloudpipeline.digital
```

PR #234, opened 2019-01-07
Add in YAML base for bosh-vars-store

LeePorte:
What
----

This was accidentally removed in rebasing.

How to review
-------------

Code review and check against https://github.com/alphagov/paas-bootstrap/commit/c66766208f3b81c0ad7dda19e1d6e5d5b869ee7a

Who can review
--------------

Anyone

### DIFF
--- a/concourse/pipelines/create-bosh-concourse.yml
+++ b/concourse/pipelines/create-bosh-concourse.yml
@@ -235,7 +235,8 @@ resources:
       bucket: ((state_bucket))
       region_name: ((aws_region))
       versioned_file: bosh-vars-store.yml
-      initial_version: "{}"
+      initial_version: "-"
+      initial_content_text: "{}"
 
   - name: bosh-init-state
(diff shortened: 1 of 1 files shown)


concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -235,7 +235,8 @@ resources:
       bucket: ((state_bucket))
       region_name: ((aws_region))
       versioned_file: bosh-vars-store.yml
-      initial_version: "{}"
+      initial_version: "-"
+      initial_content_text: "{}"
```
bandesz, line 239: I think this line is not needed, especially, that this is a yaml file and not a json

bandesz: We talked about this and there is a migration script which doesn't handle an empty yaml.

bandesz (approved):

PR #243, opened 2019-02-01
[#163558206] Use a common vcap password for all VMs

alext:
What
----

This is the first part of this. It renames the bosh_vcap_password to just vcap_password, and uses it for both the bosh and concourse VMs.

The second part will be in paas-cf to make the CF and Prometheus deployments also use this password.

As part of this I've taken the opportunity to cleanup some of the manifest generation where too much logic had escaped into the pipeline.

How to review
-------------

* Code review (might be easier commit at a time).
* Run this down a dev pipeline, and verify you can ssh to the bosh and concourse VMs (`make dev ssh_bosh` and `make dev ssh_concourse`), and then sudo using the password from bosh-secrets.yml

Who can review
--------------

Not me.

### DIFF
--- a/manifests/concourse-manifest/spec/support/manifest_helpers.rb
+++ b/manifests/concourse-manifest/spec/support/manifest_helpers.rb
@@ -9,12 +9,23 @@ module ManifestHelpers
   class Cache
     include Singleton
     attr_accessor :manifest_with_defaults
+    attr_accessor :manifest_with_github_auth
     attr_accessor :concourse_secrets_file
     attr_accessor :concourse_secrets_data
   end
 
   def manifest_with_defaults
-    Cache.instance.manifest_with_defaults ||= load_default_manifest
+    Cache.instance.manifest_with_defaults ||= render_manifest
+  end
+
+  def manifest_with_github_auth
+    Cache.instance.manifest_with_github_auth ||= render_manifest(
+      override_env: {
+        'ENABLE_GITHUB' => 'true',
+        'GITHUB_CLIENT_ID' => 'dummy_github_client_id',
+        'GITHUB_CLIENT_SECRET' => 'dummy_github_client_secret',
+      }
+    )
   end
 
   def concourse_secrets_file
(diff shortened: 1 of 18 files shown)


manifests/shared/spec/support/fixture_helpers.rb
```diff
@@ -14,6 +14,17 @@ def copy_fixture_file(file, target_dir, target_file = file)
     FileUtils.cp(fixtures_dir.join(file), "#{target_dir}/#{target_file}")
   end
 
+  def generate_bosh_secrets_fixture(target_dir)
+    FileUtils.mkdir(target_dir) unless Dir.exist?(target_dir)
+    File.open("#{target_dir}/bosh-secrets.yml", 'w') do |file|
+      output, error, status = Open3.capture3(File.expand_path("../../../bosh-manifest/scripts/generate-bosh-secrets.rb", __dir__))
+      unless status.success?
+        raise "Error generating cf-secrets, exit: #{status.exitstatus}, output:\n#{output}\n#{error}"
```
bandesz: This should say error generating bosh-secrets

manifests/concourse-manifest/spec/support/manifest_helpers.rb
```diff
@@ -29,31 +40,43 @@ def concourse_secrets_value(key)
...
-        File.expand_path("../../../concourse-base.yml", __FILE__),
-        concourse_secrets_file,
-        File.expand_path("../../../../shared/spec/fixtures/concourse-terraform-outputs.yml", __FILE__),
-        File.expand_path("../../../../shared/spec/fixtures/bosh-terraform-outputs.yml", __FILE__),
-        File.expand_path("../../../../shared/spec/fixtures/vpc-terraform-outputs.yml", __FILE__),
-      ].join(' ')
+      env,
+      root.join("manifests/concourse-manifest/scripts/generate-manifest.sh").to_s,
     )
     expect(status).to be_success, "build_manifest.sh exited #{status.exitstatus}, stderr:\n#{error}"
```
bandesz: this should say `generate-manifest.sh exited`

manifests/shared/spec/support/fixture_helpers.rb
```diff
@@ -0,0 +1,35 @@
+module FixtureHelpers
+  def terraform_fixture_value(key, fixture)
+    YAML.load_file(fixtures_dir.join("#{fixture}-terraform-outputs.yml")).fetch("terraform_outputs_#{key}")
+  end
+
+  def copy_terraform_fixtures(target_dir, fixtures = %w(vpc bosh cf))
```
bandesz: maybe we should remove the default value for the fixtures from here as we define it everywhere

bandesz (approved):
Looks good! Some minor comments to address.

The scope blew up quite much :)

manifests/shared/spec/support/fixture_helpers.rb
alext: Good spot. Fixed.

manifests/concourse-manifest/spec/support/manifest_helpers.rb
alext: Fixed

PR #244, opened 2019-02-01
[#163558206] Stop using separate ssh keypairs for bosh and concourse.

alext:
What
----

There's no need for these to have a separate key pair - it's just 
more things for the pipeline to manage, and extra points of
complexity.

This removes the bosh and concourse keypairs, and makes them use the default
keypair instead.

How to review
-------------

* Code review is probably enough.
* I've run this down my dev environment. I've also tested creating and destroying a new bosh and concourse.

Who can review
--------------

Not me.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,7 +183,7 @@ showenv: check-env-vars ## Display environment information
 	$(eval export TARGET_CONCOURSE=bootstrap)
 	@concourse/scripts/environment.sh
 	@echo export CONCOURSE_IP=$$(aws ec2 describe-instances \
-		--filters 'Name=tag:Name,Values=concourse/*' "Name=key-name,Values=${DEPLOY_ENV}_concourse_key_pair" \
+		--filters "Name=tag:deploy_env,Values=${DEPLOY_ENV}" 'Name=tag:instance_group,Values=concourse' \
 		--query 'Reservations[].Instances[].PublicIpAddress' --output text)
 	@echo export BOOTSTRAP_CONCOURSE_IP=$$(aws ec2 describe-instances \
 		--filters 'Name=tag:Name,Values=*concourse' "Name=key-name,Values=${VAGRANT_SSH_KEY_NAME}" \
--- a/manifests/concourse-manifest/concourse-base.yml
+++ b/manifests/concourse-manifest/concourse-base.yml
@@ -45,7 +45,6 @@ resource_pools:
       ephemeral_disk:
         size: 102400
         type: gp2
-      key_name: (( grab terraform_outputs_concourse_key_pair_name ))
       security_groups:
       - (( grab terraform_outputs_bosh_managed_security_group ))
       - (( grab terraform_outputs_concourse_security_group ))
--- a/scripts/bosh
+++ b/scripts/bosh
@@ -27,14 +27,14 @@ case "$AWS_ACCOUNT" in
   ;;
 esac
 
-BOSH_ID_RSA="$(aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/bosh_id_rsa" - | base64)"
+BOSH_ID_RSA="$(aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/id_rsa" - | base64)"
 export BOSH_ID_RSA
 
 BOSH_CA_CERT="$(aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/bosh-CA.crt" -)"
 export BOSH_CA_CERT
 
 BOSH_IP=$(aws ec2 describe-instances \
-	--filters "Name=key-name,Values=${DEPLOY_ENV}_bosh_ssh_key_pair" \
+	--filters "Name=tag:deploy_env,Values=${DEPLOY_ENV}" 'Name=tag:instance_group,Values=bosh' \
 	--query 'Reservations[].Instances[].PublicIpAddress' --output text)
 export BOSH_IP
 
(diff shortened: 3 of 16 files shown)


bandesz (changes requested):
In the destroy-terraform task there is a reference to the `bosh_id_rsa.pub` which I think we don't need anymore:

```
                touch paas-bootstrap/terraform/bosh/id_rsa.pub paas-bootstrap/terraform/bosh/bosh_id_rsa.pub
```

manifests/concourse-manifest/concourse-base.yml
```diff
@@ -45,7 +45,6 @@ resource_pools:
       ephemeral_disk:
         size: 102400
         type: gp2
-      key_name: (( grab terraform_outputs_concourse_key_pair_name ))
```
bandesz, line 48: Why are you removing this?

alext: Because this is reverting to use the default key. The bosh manifest specifies a default key, and that one gets used if a key_name isn't specified at the instance_group level like this. Removing this override makes it use that default.

scripts/bosh
```diff
@@ -27,14 +27,14 @@ case "$AWS_ACCOUNT" in
...
-BOSH_ID_RSA="$(aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/bosh_id_rsa" - | base64)"
+BOSH_ID_RSA="$(aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/id_rsa" - | base64)"
 export BOSH_ID_RSA
 
 BOSH_CA_CERT="$(aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/bosh-CA.crt" -)"
 export BOSH_CA_CERT
 
 BOSH_IP=$(aws ec2 describe-instances \
-	--filters "Name=key-name,Values=${DEPLOY_ENV}_bosh_ssh_key_pair" \
+	--filters 'Name=tag:Name,Values=bosh/*' "Name=key-name,Values=${DEPLOY_ENV}_key_pair" \
```
bandesz: We could use these filters instead: `"Name=tag:deploy_env,Values=${DEPLOY_ENV}" 'Name=tag:job,Values=bosh' `

Makefile
```diff
@@ -183,7 +183,7 @@ showenv: check-env-vars ## Display environment information
 	$(eval export TARGET_CONCOURSE=bootstrap)
 	@concourse/scripts/environment.sh
 	@echo export CONCOURSE_IP=$$(aws ec2 describe-instances \
-		--filters 'Name=tag:Name,Values=concourse/*' "Name=key-name,Values=${DEPLOY_ENV}_concourse_key_pair" \
+		--filters 'Name=tag:Name,Values=concourse/*' "Name=key-name,Values=${DEPLOY_ENV}_key_pair" \
```
bandesz: We could use these filters instead: `"Name=tag:deploy_env,Values=${DEPLOY_ENV}" 'Name=tag:job,Values=concourse' `

scripts/bosh-cli.sh
```diff
@@ -2,14 +2,14 @@
...
-BOSH_ID_RSA="$(aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/bosh_id_rsa" - | base64)"
+BOSH_ID_RSA="$(aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/id_rsa" - | base64)"
 export BOSH_ID_RSA
 
 BOSH_CA_CERT="$(aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/bosh-CA.crt" -)"
 export BOSH_CA_CERT
 
 BOSH_IP=$(aws ec2 describe-instances \
-    --filters "Name=key-name,Values=${DEPLOY_ENV}_bosh_ssh_key_pair" \
+    --filters 'Name=tag:Name,Values=bosh/*' "Name=key-name,Values=${DEPLOY_ENV}_key_pair" \
```
bandesz: We could use these filters instead: `"Name=tag:deploy_env,Values=${DEPLOY_ENV}" 'Name=tag:job,Values=bosh' `

scripts/ssh_bosh.sh
```diff
@@ -1,14 +1,14 @@
 #!/bin/bash
 set -euo pipefail
 
-BOSH_KEY=/tmp/bosh_id_rsa.$RANDOM
+BOSH_KEY=/tmp/id_rsa.$RANDOM
 BOSH_IP=$(aws ec2 describe-instances \
-    --filters "Name=key-name,Values=${DEPLOY_ENV}_bosh_ssh_key_pair" \
+    --filters 'Name=tag:Name,Values=bosh/*' "Name=key-name,Values=${DEPLOY_ENV}_key_pair" \
```
bandesz: We could use these filters instead: `"Name=tag:deploy_env,Values=${DEPLOY_ENV}" 'Name=tag:job,Values=bosh' `

bandesz:
There is one line we need to change in paas-cf: https://github.com/alphagov/paas-cf/blob/0d01f74d0367a541bbc4b00a1950d27856feac1c/Makefile#L211

alext:
> In the destroy-terraform task there is a reference to the bosh_id_rsa.pub which I think we don't need anymore

Doh! I got the equivalent one for concourse, but missed this one. Fixed.

bandesz:
We definitely need to change the filters, because now these return with no results.

bandesz (approved):

PR #271, opened 2019-04-26
[#164728826] Upgrade Bosh to v269.0.0

mogds:
What
----

- Upgrades Bosh from 268.4.0 to 269.0.0
- Deletes the `enable_post_deploy` removal. This was set to true in bosh.yml but removed in the ops file so it defaulted to false. Removing the the ops file removal will now set this to true
  - For more info, see https://github.com/cloudfoundry/bosh/releases/tag/v268.7.0 

How to review
-------------

- Code review
- Check that create-bosh-concourse has run successfully in a dev env (see https://deployer.towers.dev.cloudpipeline.digital/teams/main/pipelines/create-bosh-concourse)
- (optionally) Run create-bosh-concourse yourself in your own dev env

Who can review
--------------

Not @mogds or @richardTowers

### DIFF
--- a/manifests/bosh-manifest/operations.d/082-update-bpm-release.yml
+++ b/manifests/bosh-manifest/operations.d/082-update-bpm-release.yml
@@ -3,6 +3,6 @@
   path: /releases/name=bpm
   value:
     name: "bpm"
-    version: "0.13.0"
-    url: "https://bosh.io/d/github.com/cloudfoundry-incubator/bpm-release?v=0.13.0"
-    sha1: "4b6ebfdaa467c04855528172b099e565d679e0f5"
+    version: "1.0.4"
+    url: "https://bosh.io/d/github.com/cloudfoundry-incubator/bpm-release?v=1.0.4"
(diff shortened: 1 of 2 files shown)


richardTowers:
Note: the commit that went down my pipeline didn't include the `bpm` upgrade - running it again now that's been added.

richardTowers:
Latest run down my create-bosh-concourse ran with b412ee5: https://deployer.towers.dev.cloudpipeline.digital/teams/main/pipelines/create-bosh-concourse

richardTowers:
Main confusing bit of this was the `enable_post_deploy` thing.

The bosh director's bosh release used to default this to `false`, but upstream bosh deployment overrode it to `true`. We were removing the override, so it would get its default (of `false`). This release changes the default in the release to `true`, so the override in the upstream bosh deployment is now a no-op (overriding `true` to `true`), and therefore removing the override is confusing.

Net result is that previously `enable_post_deploy` was `false`, now it is `true`.

tlwr (approved):

tlwr:
I have run this down my dev environment just fine

LGTM

PR #281, opened 2019-05-14
Fix bootstrap when tagging the colocated worker

tlwr:
What
----

Since we are doing multiple workers we need to ensure that there exists
a worker with a colocated-with-web tag and one which has no tags

This commit does the same for bootstrap as we have for regular concourse

How to review
-------------

Create your concourse using this branch, you probably need to as I probably broke your Concourse

Who can review
--------------

Not @tlwr


tlwr:
I've used this to create and destroy a concourse

LeePorte (approved):
Bootstrapped from scratch and run down my stood up env

PR #307, opened 2019-09-24
Upload concourse secrets to Credhub

tlwr:
[Story](https://www.pivotaltracker.com/story/show/168427704)

_[See also](https://github.com/alphagov/paas-cf/pull/2081)_

What
----

We can use Concourse secrets defined in Credhub instead of setting them
in the pipeline; for all pipelines except create-bosh-concourse.

We cannot use Credhub for the Concourse password in
create-bosh-concourse even though by the time Concourse exists there is
a credhub, because of the bootstrapping problem. The bootstrap concourse
does not have a config server configured so we cannot use the
((concourse_web_password) variable.

We can do the same check with `try / on_error / do` at the end of the pipeline, but this time at the start of the pipeline when running `self-update-pipeline`

There was a very rarely used concourse task for rotating concourse creds, which amount to "run this then run the concourse deploy (without init bucket because that would break it). If we need to rotate the Concourse credentials we can use `credhub regenerate` instead.

How to review
-------------

Run this pipeline

Run `make dev deployer-concourse pipelines`

Spin up a bootstrap concourse

Merge it, and run `create-bosh-concourse` on:

- [ ] `ci`
- [ ] `stg-lon`
- [ ] `prod`
- [ ] `prod-lon`

Delete the `concourse-secrets.yml` from ci/stg-lon/prod/prod-lon buckets

Who can review
--------------

Not @tlwr


tlwr:
I've tested that bootstrapping a deployer concourse works and that running it on an existing one also works

concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -345,25 +336,104 @@ jobs:
...
+                path: sh
+                args:
+                  - -e
+                  - -c
+                  - |
+                    echo "${TARGET_CONCOURSE}"
+
+                    if [ "$TARGET_CONCOURSE" = "bootstrap" ]; then
+                      echo This concourse is bootstrap, passing this task
+                      echo pass > check-if-bootstrap/concourse.txt
```
richardTowers: I think the `check-if-bootstrap` name isn't descriptive enough of what this is being used for. It's actually more like "ignore-self-update-pipelines-failure" or something.

tlwr: 41a5fbe6a81976514900ca70ba583276a55c3219

tlwr:
I neglected to think about the case for an existing Concourse where we aren't setting the pipeline from computer.

In that case we run `make <env> pipelines` from master and it will work. I can do this at release time lest we add another conditional

richardTowers:
This went green in my existing env, and I (eventually) managed to set the right combination of env vars to get it to work in a bootstrap concourse too (BRANCH and GITHUB_CLIENT_ID/SECRET).

Just destroying the bootstrap env to check that works, then I think this is ready to roll.

For each environment, we'll need to:

* Pause create-cloudfoundry
* Run `make $env deployer-concourse pipelines SELF_UPDATE_PIPELINE=false`
* Run create-bosh-concourse
* Run `make $env deployer-concourse pipelines SELF_UPDATE_PIPELINE=true`
* Run create-bosh-concourse again just to be sure

We'll need to tell all the developers to do this too, or they'll get a nasty surprise.

PR #313, opened 2019-10-04
Add some documentation on github access

richardTowers:
What
----

This is on by default now, so if these variables aren't set when you
bootstrap your environment it won't work.

How to review
-------------

* Docs review is enough

Who can review
--------------

Not @richardtowers

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,22 @@ $ export DEPLOY_ENV=environment-name
 
 It is important that you do not use the same `DEPLOY_ENV` for both build and deployer concourse environments, this is to avoid conflicts in resource allocation.
 
+* Choose whether to enable GitHub access and if so, set `GITHUB_CLIENT_ID` and `GITHUB_CLIENT_SECRET`
+
+By default, the developers on the team can use GitHub to log in to your dev environment. To make this work you need to
+set up a GitHub OAuth application, as [described in the team manual](https://team-manual.cloud.service.gov.uk/guides/Github_oAuth_in-Dev/)
+
+```
+$ export GITHUB_CLIENT_ID=some-github-client-id
+$ export GITHUB_CLIENT_SECRET=some-github-client-secret
+```
+
+If you'd prefer not to enable GitHub single-sign-on then you can disable it with:
+
+```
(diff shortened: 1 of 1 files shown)


LeePorte (approved):

README.md
```diff
@@ -70,6 +70,22 @@ $ export DEPLOY_ENV=environment-name
...
+$ export GITHUB_CLIENT_ID=some-github-client-id
+$ export GITHUB_CLIENT_SECRET=some-github-client-secret
+```
+
+If you'd prefer not to enable GitHub single-sign-on then you can disable it with:
+
+```
+$ export ENABLE_GITHUB=false
+```
+
```
tlwr, line 88:
```suggestion

You can permanently enable Github by using `make dev upload-github-oauth` which looks in `paas-credentials/github.com/concourse/dev/client_id.gpg` and `paas-credentials/github.com/concourse/dev/client_secret.gpg` for the required Github credentials.
```

tlwr:
Github didn't refresh :(

README.md
richardTowers: Does this actually work though? How does GitHub know the redirect URI to use if we're all using the same oauth token?

tlwr: `paas-credentials` doesn't have a dev secret, you're expected to put it their yourself and not commit it.

The make task which puts it in s3 puts it in your DEPLOY_ENV's state bucket

tlwr: See https://github.com/alphagov/paas-bootstrap/blob/master/scripts/manage-github-secrets.sh#L6

It is then pulled in when setting the pipeline, which first tries s3 then tries paas-credentials

richardTowers: Ah yeah, I knew about that weird workflow. I don't like it though.

Setting them as env vars the first time will already set them permanently as they'll be in credhub. Why bother storing them locally too, since they're only used for this one feature.

tlwr: These variables, to my knowledge, are not stored in Credhub, because we want them to be persisted in developer environments after `bootstrap-destroy`. I can't see in the code where we put them in Credhub.

PR #323, opened 2019-11-14
Update UAA to 74.7.0

richardTowers:
What
----

Updated UAA to v74.7.0 to match paas-cf and patch security issues.

How to review
-------------

* Check this matches https://bosh.io/releases/github.com/cloudfoundry/uaa-release?version=74.7.0
* Check that someone has run this down create-bosh-concourse and hasn't
  bricked their environment (I'm doing this now)

Who can review
--------------

Not @richardtowers, probably best for @46Bit to look at it.


46bit (approved):

richardTowers:
create-bosh-concourse ran fine with this in my dev env.

46bit:
that's good because I already merged it 😅

PR #329, opened 2019-12-06
[#169915408] Use Google SSO to authenticate to bosh's UAA

richardTowers:
What
----

We don't want to be using shared credentials to access bosh, as this
makes auditing impossible. We also don't want to have to manage another
set of credentials, so it's best if we can use SSO.

This adds the bosh-admin users from [alphagov/paas-trusted-people](https://github.com/alphagov/paas-trusted-people/) to
bosh's UAA (via its config file), and updates the bosh-cli script to
allow people to log in using SSO.

I also did a bit of updating of things on Travis because I was annoyed
at them being much older versions than I had locally.

See the individual commits for more detail.

How to review
-------------

* Code review
* Run this branch down your dev env
* Check that you can authenticate to bosh in you dev env using `make dev bosh-cli` from this branch

Who can review
--------------

Not @richardtowers

### DIFF
--- a/concourse/pipelines/create-bosh-concourse.yml
+++ b/concourse/pipelines/create-bosh-concourse.yml
@@ -158,6 +158,15 @@ resources:
       initial_version: "-"
       initial_content_text: ""
 
+  - name: bosh-uaa-google-oauth-secrets
+    type: s3-iam
+    source:
+      bucket: ((state_bucket))
+      region_name: ((aws_region))
+      versioned_file: bosh-uaa-google-oauth-secrets.yml
+      initial_version: "-"
+      initial_content_text: ""
+
   - name: bootstrap-cyber-tfvars
     type: s3-iam
     source:
@@ -809,6 +818,7 @@ jobs:
           passed: ['bosh-terraform']
         - get: bosh-tfstate
           passed: ['bosh-terraform']
+        - get: bosh-uaa-google-oauth-secrets
 
       - task: extract_terraform_outputs
         config:
@@ -843,6 +853,7 @@ jobs:
             - name: bosh-secrets
             - name: bosh-CA
             - name: bosh-vars-store
+            - name: bosh-uaa-google-oauth-secrets
           outputs:
             - name: bosh-manifest
             - name: bosh-manifest-pre-vars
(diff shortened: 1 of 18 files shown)


concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -86,6 +86,13 @@ resources:
       branch: ((branch_name))
       commit_verification_keys: ((gpg_public_keys))
 
+  - name: paas-trusted-people
+    type: git
+    source:
+      uri: https://github.com/alphagov/paas-trusted-people.git
+      branch: master
+      commit_verification_keys: ((gpg_public_keys))
```
richardTowers: 🤔  this should have a tag filter, otherwise changes will go straight to prod.

richardTowers: Although paas-bootstrap doesn't auto-trigger, so less of an issue, but still.

LeePorte: I think we should use a tag here

LeePorte (commented):
One small change so far

scripts/bosh-cli.sh
```diff
@@ -1,53 +1,114 @@
...
+pacserverpid="$!"
 
+if hash networksetup &> /dev/null && [[ "$pac_url" != "$old_auto_proxy_url" || "$old_auto_proxy_enabled" == "No" ]]; then
+  echo
+  echo 'Your system network settings need to be changed to allow your browser to access UAA over our SSH tunnel.'
+  echo 'Your current settings are: '
+  echo "$old_auto_proxy_settings"
+  echo
+  while true; do
+    read -r -p 'Would you like these to be changed? ' yn
```
LeePorte: Should include the possible responses in the prompt `Would you like these to be changed? (y/n)`

concourse/pipelines/create-bosh-concourse.yml
richardTowers: This is a moot point now that it's a private repo - the PR's been updated so the YAML file in paas-trusted people is uploaded to S3 instead of pulled from the repo by the pipeline.

scripts/bosh-cli.sh
richardTowers: Done

richardTowers:
Closing, to revisit in the new year.

PR #324, opened 2019-11-20
[#166500017] Create individual accounts

mogds:
What
----

This creates individual unix accounts with public keys provided via the os-config release. 
- Adds an ops file to add users to BOSH director
- Adds a runtime config to add users to all other VMs managed by BOSH
- Also updates ssh scripts to use individual keys. Related PR (https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/178)

How to review
-------------

- Code review
- Deploy to your dev env
- Review and merge https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/178
- Update tag in scripts/bosh-cli.sh and add any missing users

Who can review
--------------

Not me

PR #336, opened 2020-01-07
Refactor concourse/scripts/ssh.sh

46bit:
What
----

In #324 we stopped using a global SSH key and switched to user-specific SSH keys. But we didn't stop downloading the old global key from S3. This commit removes that old code.

How to review
-------------

* Code review;
* I've checked this still works but feel free to yourself.

Who can review
--------------

Not @46bit.

### DIFF
--- a/concourse/scripts/get_ip_address.sh
+++ b/concourse/scripts/get_ip_address.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+set -euo pipefail
+
+case ${1:-} in
+  concourse)
+    env > /tmp/debug.txt
+    aws ec2 describe-instances \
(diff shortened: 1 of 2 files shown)


richardTowers:
Looks like you're missing `get_ip_address.sh` - also does concourse not have a DNS entry for it's IP address? With bosh we've got `bosh.((system_domain))` and `bosh-external.((system_domain))`, so there's no need to get the IP address, you can just use the external hostname.

46bit:
@richardTowers A `concourse-external.((system_domain))` would be lovely but from a quick look we don't have it, only an ELB domain which won't work for SSH. I'm not prepared to add that right now–this PR only exists to remove an invokation of `make` so that the CLI can work.

PR #337, opened 2020-01-08
Use default SSH key when SSHing to Concourse

46bit:
What
----

Some of the team use SSH keys on disk to access Concourse, but others use keys on a Yubikey. The existing script didn't support using a Yubikey.

This commit removes all specifying of the SSH key, relying on the default. It also removes `-o IdentitiesOnly=yes` which was obstructing Yubikey use.

How to review
-------------

Get two people to review: one who uses a Yubikey and one who uses a key on disk. Check it works for both of you.

Who can review
--------------

Not @46bit.

### DIFF
--- a/concourse/scripts/ssh.sh
+++ b/concourse/scripts/ssh.sh
@@ -2,8 +2,6 @@
 
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


tlwr:
Works for me using disk based `id_rsa`

PR #342, opened 2020-01-29
Use default SSH key when SSHing to BOSH

46bit:
What
----

Some of the team use SSH keys on disk to access BOSH, but others use keys on a Yubikey. The existing script didn't support using a Yubikey.

This commit removes all specifying of the SSH key, relying on the default. It also removes `-o IdentitiesOnly=yes` which was obstructing Yubikey use.

How to review
-------------

Someone who uses a key on disk should review this. You may as well do it alongside #337 which is the Concourse version of this PR.

Who can review
--------------

Not @46bit

### DIFF
--- a/scripts/ssh_bosh.sh
+++ b/scripts/ssh_bosh.sh
@@ -1,16 +1,9 @@
 #!/bin/bash
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


tlwr (approved):
Works for me using disk based `id_rsa`

PR #349, opened 2020-02-18
[#169623326] Ship Concourse Auth Logs to Splunk

mogds:
What
----

Ships concourse's auth.log to splunk using the awslogs release.

How to review
-------------

Code review and deploy to your dev env 

Who can review
--------------

Not me

### DIFF
--- a/manifests/concourse-manifest/concourse-base.yml
+++ b/manifests/concourse-manifest/concourse-base.yml
@@ -139,22 +139,6 @@ instance_groups:
       - name: concourse
         default: [dns, gateway]
 
-addons:
-- name: awslogs
-  jobs:
-  - name: awslogs-xenial
-    release: awslogs
-    properties:
-      awslogs-xenial:
-        region: (( grab $AWS_REGION ))
-        awslogs_files_config:
-          - name: /var/log/auth.log
-            file: /var/log/auth.log
-            log_group_name: (( concat "concourse_auth_logs_" terraform_outputs_environment ))
-            log_stream_name: "{{instance_id}}"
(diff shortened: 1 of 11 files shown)


terraform/cyber/cloudwatch_logs.tf
```diff
@@ -1,6 +1,6 @@
 locals {
   destination_arn = "${replace(var.csls_kinesis_destination_arn, "REGION", var.region)}"
-  log_groups      = "${formatlist("%s_%s", var.log_groups_to_ship_to_csls, var.env)}"
+  log_groups      = "${formatlist("%s_%s", concat(var.log_groups_to_ship_to_csls, list("concourse_auth_logs")), var.env)}"
```
tlwr: Can we add `"concourse_auth_logs"` to `terraform/globals.tf` instead of this `concat(...)`

Or refactor the terraform so that there are `bosh_log_groups_to_ship_to_csls` and `concourse_log_groups_to_ship_to_csls` variables and then refactor `terraform/concourse/cloudwatch_logs.tf` to be more similar to `terraform/bosh/cloudwatch_logs.tf`

manifests/concourse-manifest/concourse-base.yml
```diff
@@ -47,6 +51,19 @@ instance_groups:
         release: bpm
         properties: {}
 
+      - name: awslogs-xenial
```
tlwr: This is only one instance group, can we make this a [BOSH deployment manifest addon](https://bosh.io/docs/manifest-v2/#addons) instead so we also get logs from the workers

PR #355, opened 2020-03-03
Add Seb's public SSH key

paroxp:
How to review
-------------

Check with Seb if both username and public key are correct.


schmie (approved):
cheers :+1:

PR #358, opened 2020-03-17
[#171787442] Bosh access UAA via ALB

tlwr:
[Story](https://www.pivotaltracker.com/n/projects/1275640/stories/171787442)

What
----

Adds an ALB in front of UAA running on the BOSH director so we can access UAA using a publicly valid certificate

ALB is admin CIDR restricted

UAA is addressable on `bosh-uaa-external.((system_domain))` so it is obvious that it is external, and related to BOSH.

Fixup the `destroy-bosh-cloudfoundry` pipeline to include cyber terraform, so that it works.

Gotchas
----

- [this issue about terraform AWS](https://github.com/terraform-providers/terraform-provider-aws/issues/7617) is mentioned in the story that allegedly prevents the attachment of an IP to a target group, I did not encounter this when testing
- In the UAA section of the BOSH manifest, UAA needs to know its external URL. The director SHOULD NOT know about UAA's external URL

How to review
-------------

Code review

`make dev bosh-cli DEPLOY_ENV=tlwr`

`echo | openssl s_client -connect deployer.tlwr.dev.cloudpipeline.digital:443 | openssl x509 -issuer -noout` should only list AWS related certs

`make dev bosh-cli` should print the correct URL in the prelude

Try it in your dev env

Who can review
--------------

Not @tlwr

### DIFF
--- a/concourse/pipelines/destroy-bosh-concourse.yml
+++ b/concourse/pipelines/destroy-bosh-concourse.yml
@@ -76,6 +76,37 @@ resources:
       region_name: ((aws_region))
       versioned_file: bosh.tfstate
 
+  - name: bootstrap-cyber-tfvars
+    type: s3-iam
+    source:
+      bucket: ((state_bucket))
+      region_name: ((aws_region))
+      versioned_file: bootstrap-cyber.tfvars
+      initial_version: "-"
+      initial_content_text: ""
+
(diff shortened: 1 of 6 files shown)


paroxp (approved):
Tested on tlwr's environment.

paroxp:
Tested in my dev env fine.

PR #361, opened 2020-04-06
Upgrade concourse to 6.0.0

LeePorte:
What
----

We need to keep up with the times, so this upgrades concourse to the latest version

How to review
-------------

Code review and deploy to dev (2 runs of create-bosh-concourse are required)

Who can review
--------------

Anyone

### DIFF
--- a/manifests/concourse-manifest/concourse-base.yml
+++ b/manifests/concourse-manifest/concourse-base.yml
@@ -16,9 +16,9 @@ stemcells:
 # relevant tag of https://github.com/concourse/concourse-bosh-deployment
 releases:
   - name: "concourse"
-    version: "5.8.0"
-    url: "https://bosh.io/d/github.com/concourse/concourse-bosh-release?v=5.8.0"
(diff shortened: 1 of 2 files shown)


tlwr (approved):
Having it stop at `concourse-deploy` and then starting `init-pipeline` worked for me

paroxp (approved):
Looks good in general.

There is one problem I'm noticing...

If I try to start the job somewhere in the middle of the pipeline, (the original trigger not the new one) it will not start...

It does however work fine, running from beginning. Change of habit I suppose. Would either of you attempt to replicate that issue and decide if it's legit blocker or not?

tlwr:
I think this is an intended change in behaviour in Concourse and you have to use the new "trigger with same inputs" button - or pin the resource

LeePorte:
Correct, this is a change in behaviour with the new algorithm introduced https://github.com/concourse/concourse/releases/tag/v6.0.0

paroxp:
Figured. Just so we're aware :) Merge away!

PR #375, opened 2020-07-01
Shrink deployer-concourse to m5.large in dev

philandstuff:
Each team member runs an always-on deployer concourse consisting of a
web node and a worker node, each of which is an m5.xlarge instance.
If each m5.xlarge costs $0.222/hour (current london price; ireland is
slightly cheaper), this corresponds to an on-demand cost of
$0.222×2×24×30 = $319.68 per 30-day month, per developer concourse.

By shrinking the instance size to m5.large we could save half of this
cost.

What
----

Describe what you have changed and why.

How to review
-------------

Describe the steps required to test the changes.

Who can review
--------------

Describe who can review the changes. Or more importantly, list the people
that can't review, because they worked on it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,7 @@ dev: globals check-env-vars ## Set Environment to DEV
 	$(eval export SKIP_COMMIT_VERIFICATION=true)
 	$(eval export AWS_DEFAULT_REGION ?= eu-west-1)
 	$(eval export CYBER_PASSWORD_STORE_DIR?=${HOME}/.paas-pass)
+	$(eval export CONCOURSE_INSTANCE_TYPE=m5.large)
 
 .PHONY: ci
 ci: globals check-env-vars ## Set Environment to CI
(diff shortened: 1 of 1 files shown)


philandstuff:
oops, stupid PR templates going to the bottom of the text box and so being below the fold if you have a sensible-sized commit message

tlwr:
Cherrypicked to https://github.com/alphagov/paas-bootstrap/tree/wip-tlwr and running through my dev env

tlwr:
I have deployed this to my development environment and run the pipeline a couple of times

We see an increased load average compared to previous days, this isn't necessarily a bad thing, it just means we're actually the instance proportionally, as this is only in dev, I am going to merge this.

The first peak is the instance provisioning, the second two peaks are the tests being run

image

tlwr:
Also whilst doing this firebreak work, Phil awared me Towers and TNW still having Concourses lying around, which I've deleted

PR #381, opened 2020-08-03
Try to fix Cyber CSLS Kinesis Destination ARN

46bit:
What
----

I'm trying to reset (destroy + recreate) my dev env. Running Terraform from the bootstrap Concourse is failing with:

```
Error: Required variable not set: csls_kinesis_destination_arn
```

That `csls_kinesis_destination_arn` secret was not being uploaded anywhere in `paas-bootstrap`. Meanwhile, `paas-cf` only uploads it to Credhub, which doesn't exist at bootstrap time.

This commit gets `paas-bootstrap` to upload `csls_kinesis_destination_arn` to S3 to fix the problem.

How to review
-------------

Code review. Sanity check we're OK putting this in S3.

Who can review
--------------

Not @46bit.

### DIFF
--- a/manifests/bosh-manifest/scripts/generate-manifest.sh
+++ b/manifests/bosh-manifest/scripts/generate-manifest.sh
@@ -103,6 +103,7 @@ admin_google_oauth_client_id: ((admin_google_oauth_client_id))
 admin_google_oauth_client_secret: ((admin_google_oauth_client_secret))
 
 bosh_auditor_splunk_hec_token: ((bosh_auditor_splunk_hec_token))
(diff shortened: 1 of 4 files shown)


AP-Hunt (approved):
This looks like a solid change to me

46bit:
Unfortunately not working yet. :(

46bit:
To the poor unfortunate who comes across this pull request in future, this problem stops after you run the `upload-cyber-tfvars` Make target. I would improve the error myself but I'm exhausted from stopping my bootstrap deploying to the wrong VPC (aaa.)

PR #386, opened 2020-08-24
Update pipelines as the first step

LeePorte:
What
----

Recently it was discovered that we were performing other actions before
updating the pipeline, this caused an issue where the docker container
that contained terraform had been updated but a previous version was
being used as it was required before the update step.

How to review
-------------

Code review

Who can review
--------------

Anyone


barsutka:
Code review: looks good to me.

PR #392, opened 2020-09-14
Terraform 0.13

LeePorte:
What
----

Updated our terraform to use version 0.13.2 as it is the most current release

How to review
-------------

Run it down your pipeline

Who can review
--------------

Anyone

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ rvm:
   - 2.7
 env:
   global:
-    - TF_VERSION="0.12.29"
+    - TF_VERSION="0.13.2"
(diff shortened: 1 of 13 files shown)


AP-Hunt (commented):
You might need to set [an extra version constraint here](https://github.com/alphagov/paas-bootstrap/blob/terraform-0.13/terraform/providers.tf#L6) for the `codecommit` alias of the AWS provider

paroxp (approved):
Checked on Lee's environment, with the only change of aws provider path being updated.

WIP commit to be fixed after approval.

PR #397, opened 2020-10-06
sync autom8

philandstuff:
What
----

Update autom8 membership

.o(is it possible to get concourse to use the autom8 github team
directly rather than maintaining separate lists?)


chrisfarms (approved):

paroxp:
You could be using the GitHub team if you rather.

I think the schema is based on this: https://concourse-ci.org/github-auth.html#github-authorization

paroxp:
The granularity is useful to manage SC based roles, where as we still want people to access repos belonging to the team.

PR #400, opened 2020-10-12
Ignore RDS engine versions using lifecycle rule

tlwr:
What
----

We set `auto_minor_version_upgrade` on our RDS instances, which means that their minor versions get upgraded, this most recently happened from `11.5` to `11.8`.

However our maintenance windows are not synchronised, so there may be 24h or so when we have inconsistent versions.

We can ignore the engine version using a lifecycle rule, and let AWS manage the version. This lifecycle rule can be removed when we are explicitly upgrading the versions.

How to review
-------------

Code review


46bit (approved):
Approved as basically harmless to keep you unblocked

PR #409, opened 2020-11-10
Bootstrap docker-compose concourse has correct versions

tlwr:
What
----

When I was destroying my bosh-concourse via bootstrap, I noticed that Concourse was being stood up with 5.8 and 6.6 concurrently.

To remedy this I:
- added rspec to the `vagrant` directory
- wrote a test which checks that all similar images use an identical version (eg `concourse:6.5` and `concourse:6.6` fails)
- updated `docker-compose.yml`

Additionally, `Waiting for concourse to start for 180s` was timing out, because `localhost` could not be resolved temporarily (Mac OS shenanigans I imagine). I have remedied this by using loopback instead (`127.0.0.1`)

How to review
-------------

CI

Code review


schmie (approved):
:+1:

PR #424, opened 2021-01-22
Update to use `main`

LeePorte:
What
----

Switch over to main as the default branch

How to review
-------------
code review

Who can review
--------------

Anyone


schmie (approved):
:+1: confirmed with @LeePorte that this requires changes in other places to not break things.
As bootstrap is run infrequently it is a good place to start.

PR #445, opened 2021-04-09
We don't use postgres 9.5 for bosh any longer

LeePorte:
What
----

As we don't use postgres 9.5 with BOSH there is no point in having a parameter group for
something we don't use.

How to review
-------------

Code review

Who can review
--------------

Anyone

---


AP-Hunt (approved):
Lee has shown that there are no databases using the parameter group any more.

PR #459, opened 2021-06-17
[#178486319] concourse db: make instance_class configurable, default to t3.large (for dev)

risicle:
What
----

`db.m5.xlarge` is a bit excessive (read: expensive) for dev environments, especially given this part of the deployment doesn't get automatically spun down.

How to review
-------------

Look at it, or maybe even try it.

I tested this by spinning up my concourse-lite with `BRANCH` pointing at this branch. It might be possible to do it without having to bring your concourse-lite up, setting your main concourse's `create-bosh-concourse` directly, but I couldn't see an easy way of doing this. Please let me know if there is one.

Who can review
--------------

Anyone

---


Krenair (commented):
Do we want to also add `concourse_db_instance_class          = "db.m5.xlarge"` to terraform/ci.tfvars?

AP-Hunt:
> Do we want to also add `concourse_db_instance_class = "db.m5.xlarge"` to terraform/ci.tfvars?

Yes, we probably do. Otherwise Terraform will try to delete and re-create the database .. I think?

Krenair:
> > Do we want to also add `concourse_db_instance_class = "db.m5.xlarge"` to terraform/ci.tfvars?
> 
> Yes, we probably do. Otherwise Terraform will try to delete and re-create the database .. I think?

Well it should try to modify the existing instance. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html documents it as 'Some interruptions' (not Replacement) and I'd expect Terraform to behave the same.

In any case I'm not sure we want to reduce the instance class of the CI Concourse DB?

risicle:
RDS instance type up/downgrades are smooth, but yes I've re-upgraded ci's db again now.

Note though that `db.m5.xlarge` is one level higher than we use for Big Concourse, which runs about 10 projects.

Krenair (approved):

PR #481, opened 2021-11-24
Update dev* make targets and pick correct Github OAuth credentials

cadmiumcat:
What
----

1. Rather than hardcoding "devxx" as targets, use a wildcard rule. From https://github.com/alphagov/paas-cf/pull/2780
2. Ensure correct Github OAuth credentials are picked for shared dev environments when uploading secrets. 

Implementation note:

Previously, `SYSTEM_DNS_ZONE_NAME` and `APPS_DNS_ZONE_NAME` were set
twice, once in `dev:` and once in the individual environment target (ie.
`dev01:`).
By escaping the `$` for the DEPLOY_ENV variable in `dev:`, it is not
expanded at the time the make target runs, but instead by the shell
at runtime. This allows a little bit of deduplication!

How to review
-------------

Code review, commit by commit.  

Who can review
--------------

Anyone but Andy
---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,9 +81,9 @@ globals:
 
 .PHONY: dev
 dev: globals check-env-vars ## Set Environment to DEV
-	$(eval export SYSTEM_DNS_ZONE_NAME=${DEPLOY_ENV}.dev.cloudpipeline.digital)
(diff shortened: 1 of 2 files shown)


risicle (approved):
I *think* this is right.

cadmiumcat:
> I _think_ this is right.

That's the spirit 😆

PR #486, opened 2022-02-15
Automatically rotate BOSH leaf certificates

risicle:
What
----
https://www.pivotaltracker.com/story/show/178640681

We decided to integrate the rotation process with the release pipeline in a similar way to the `create-cloudfoundry` pipeline's certificate handling rather than having some time-based trigger asynchronously causing the certificate update to happen behind the scenes. This does mean that it relies on the pipeline being run _during_ the expiry period for the rotation to happen in time, which is a concern because the `create-bosh-concourse` pipeline is run so much less frequently than `create-cloudfoundry`. To this end I increased the grace period to 180 days before the certificate expiry to give us a better chance.

Because all we're rotating here are the leaf certificates, I realized we don't have to do any clever two-phase rollouts for the rotation - the old certificates aren't being made invalid (they remain valid until their expiry date). So simply inspecting the certificates _before_ rollout and replacing certificates then (if deemed necessary) will suffice.

The existing `check-certificates` job was split to a separate script and given the ability to ignore CA certificates. This is then used to detect certificates close to expiry. As noted in the comments it is not sophisticated enough to rotate just those certificates that need it, but I suspect all of our leaf certificates share the same timestamp anyway.

This also does a bit of refactoring of the `create-bosh-concourse` pipeline's bosh credential handling which I found quite puzzling initially. Hopefully the end result is a little easier to understand than before. In particular, the script `rotate-vars-store-secrets.rb` was renamed to `drop-vars-store-secrets-for-rotation.rb` (along with its associated jobs & tasks) because it _doesn't actually_ rotate the secrets. It simply drops them (or possibly moves them to an `_old` sibling key) so that the next call to `generate-manifest.rb` will do the regeneration.

How to review
-------------

Deploy to a dev env and perform a deployment. If your certificates aren't due for regeneration, you can hack a `--min-remaining-days 400` flag into the `check-certificates.rb` call in the `check-bosh-leaf-certs` task, which will _always_ consider the certificates to be due for renewal.

Probably best to also check the standalone `drop-bosh-leaf-certs-for-rotation` and `check-certificates` jobs (visible in the `credentials` group) work for you.

Who can review
--------------
Not @risicle, not @milesb-gds, possibly not @cadmiumcat .

---

### DIFF
--- a/concourse/pipelines/create-bosh-concourse.yml
+++ b/concourse/pipelines/create-bosh-concourse.yml
@@ -2251,6 +2251,7 @@ jobs:
     plan:
 
     - in_parallel:
+      - get: paas-bootstrap
       - get: bosh-manifest-pre-vars
       - get: bosh-vars-store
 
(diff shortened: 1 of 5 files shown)


manifests/shared/scripts/check-certificates.rb
```diff
@@ -1,42 +1,58 @@
...
+def find_certificates(yaml, path, skip_ca_certs)
   return [] if yaml.is_a? Numeric
   return [] if yaml.is_a? TrueClass
   return [] if yaml.is_a? FalseClass
 
   if yaml.is_a? String
     return yaml.scan(CERT_REGEX).map do |cert_str|
+      cert = OpenSSL::X509::Certificate.new(cert_str)
+
+      return [] if skip_ca_certs && basic_constraints_value(cert) != "CA:FALSE"
```
AP-Hunt: Extract the second condition to a function so it has a meaningful name (or otherwise a variable). I found it hard to parse without it.

manifests/shared/scripts/check-certificates.rb
```diff
@@ -76,33 +92,31 @@ def find_certificates(yaml, path)
 puts "Found #{expd_manifest_certs.length} expired certs:"
 expd_manifest_certs.group_by(&:filepath).each do |filepath, certs|
```
AP-Hunt: `exp` and `expd` are confusingly similar. Suggest changing to "expiring" and "expired"

concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -974,63 +974,115 @@ jobs:
...
+                - |
+                  # this isn't particularly sophisticated - *any* leaf certificate
+                  # being due for rotation will cause them *all* to be regenerated
+                  ./paas-bootstrap/manifests/shared/scripts/check-certificates.rb \
+                    --skip-ca-certs \
+                    bosh-vars-store-cleaned/bosh-vars-store.yml
+
+                  # a no-op "rotation" for the successful case
+                  cp bosh-vars-store-cleaned/bosh-vars-store.yml bosh-vars-store-rotated/bosh-vars-store.yml
+          on_failure:
```
AP-Hunt: Can you explain this setup? It looks to me like it's expecting a non-zero exit code before it rotates the certificates?

AP-Hunt (changes requested):

concourse/pipelines/create-bosh-concourse.yml
risicle: Yeah I should add a comment. `check-certificates.rb` returns nonzero if it finds any expired certs.

risicle: better?

AP-Hunt: I think I'm still a bit confused why we're setting it up like a try .. catch. Why not

```
if ! ./paas-bootstrap/manifests/shared/scripts/check-certificates.rb --skip-ca-certs bosh-vars-store-cleaned/bosh-vars-store.yml; then
    # Rotate vars
fi
```

Or something to that effect. If I saw that job failing in concourse, I wouldn't assume it was supposed to. I would assume it was broken. I'm not saying we definitely should change it, just that I don't yet understand the rationale for it as it is.

risicle: :thinking: that's a good point. I can't remember why I did it like this... it might have just been that I was intending to make the `check-certificates` task a reusable task shared with the standalone `check-certificates` job. But clearly that won't work because it uses different options. So yeah I'll look into it when I get time...

risicle: Better?

AP-Hunt: Yes, much better I think

AP-Hunt (approved):

PR #492, opened 2022-03-28
[FIREBREAK] Update the ssh command for 'ssh_bosh'

whi-tw:
What
----

Before, we were ssh'ing directly to the bosh ip, which is difficult to set up ssh_config for.

This way, we ssh to `paas_bosh_ssh', but set the hostname to the IP we are ssh'ing to.

~/.ssh/config can then be set up for this fake hostname.
```
Host paas_bosh_ssh
	User xxxx
	IdentityFile %d/.ssh/xyz.pub
```

Specifying the user for the SSH to `${USER}` is unnecessary, as this is the default behaviour of SSH, and makes it impossible to override with ssh_config. Also, as `$USER` is not immutable, it's not a security improvement.

This is an improvement which allows for cases where there are a large number of ssh keys in the agent ie. if using [1Password's agent](https://developer.1password.com/docs/ssh/).
In this case, if the needed key is too far down the list, the attempts with invalid keys leads to an authfail before the correct one is tried.

How to review
-------------

1. Run `gds aws $role -- make ssh_bosh DEPLOY_ENV=$env` before and after checking out this branch, and ensure there are no changes.
1. Update your ssh config, as-in my example, and then run the `make` command again. You should see an error, as the username / key are incorrect.
1. Remove the ssh config again, and note that SSH'ing works again.

Who can review
--------------

Not me

---

### DIFF
--- a/scripts/bosh-cli.sh
+++ b/scripts/bosh-cli.sh
@@ -3,7 +3,7 @@ set -euo pipefail
 
 tunnel_mux='/tmp/bosh-ssh-tunnel.mux'
 
-function cleanup () {
+function cleanup() {
   echo 'Closing SSH tunnel'
   ssh -S "$tunnel_mux" -O exit a-destination &>/dev/null || true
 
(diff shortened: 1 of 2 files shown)


AP-Hunt:
I'd be interested in talking about this at a knowledge share session.  I think I like the flexibility it would allow, but I'm not yet sure what problems it would introduce for engineers; especially new engineers being onboarded.

whi-tw:
> I'd be interested in talking about this at a knowledge share session. I think I like the flexibility it would allow, but I'm not yet sure what problems it would introduce for engineers; especially new engineers being onboarded.

Sure, makes perfect sense.

In essence, though, there would be 0 change to how it works (ssh already uses `$USER` by default, and it doesn't matter what the 'target' is (ie. ssh **target**) if `-o Hostname=some_ip` is set). In theory, you can check this out and test it - it will *Just Work™️*, with no external changes required.

Context: without this change, I can't run this make target without running two ssh agents:

```
ssh-add -L | wc -l
      25
```

scripts/bosh-cli.sh
```diff
@@ -3,7 +3,7 @@ set -euo pipefail
 
 tunnel_mux='/tmp/bosh-ssh-tunnel.mux'
 
-function cleanup () {
```
AP-Hunt, line 6: Small check: will this cause a problem in different shells? e.g `bash` vs `zsh`?

AP-Hunt (approved):

scripts/bosh-cli.sh
whi-tw: This change was suggested by `shellcheck` and is still working on `zsh`, so I don't think so!

PR #498, opened 2022-04-26
Scale up the BOSH director prod database to db.m5.large

schmie:
What
----

We recently encountered lock-up of the BOSH director with the immediate symptom being loss of metrics. In addition to the loss of metrics we see max load on the BOSH director and high load, swap and max connections on the BOSH RDS database.
We hypothesize that the lock-up is caused by database latency in the first place.
Also abstracting bosh_db_instance_class with a default of db.t3.small. The default is in effect for all non production environments.

How to review
-------------

- Code review
- Potentially test in dev env (Would require to update dev BOSH db to HA and extend scaling to dev, so not sure about the work to gain ratio?!) 

Who can review
--------------

Not @schmie 

---

### DIFF
--- a/terraform/bosh/rds.tf
+++ b/terraform/bosh/rds.tf
@@ -41,7 +41,7 @@ resource "aws_db_instance" "bosh" {
   storage_type               = "gp2"
   engine                     = "postgres"
   engine_version             = "11.5"
-  instance_class             = "db.t3.small"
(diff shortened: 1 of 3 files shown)


AP-Hunt (approved):
I think it'd be more work than it's worth to test in in dev. It's a fairly routine database instance class change. AWS RDS should handle it cleanly.

PR #532, opened 2022-10-26
Fix pipelines.sh scripts interaction with manage-github-secrets.sh

whi-tw:
## What

The pipelines.sh calls manage-github-scripts.sh file with an eval. The eval is necessary to load the environment variables from manage-github-secrets.

It looks like this interaction was broken by this [change](https://github.com/alphagov/paas-bootstrap/commit/2966888fbcb80a2d514feb13b570a60be98dedba). The change added echo's which are also eval'd by pipelines.sh and obviously don't parse. This problem would only show up if you were running the script against an environment without secrets already in the bucket.

This updates the echos to output to stderr and splits the eval and script to ensure we exit with a failure if manage-github-secrets fails.

## How to review

This was tested on the command line while creating the vagrant box.
Check you are happy with the way this has been fixed.

## Who can review

One of the old timers

---

### DIFF
--- a/concourse/scripts/pipelines.sh
+++ b/concourse/scripts/pipelines.sh
@@ -63,7 +63,8 @@ EOF
 }
 
 if [ "${ENABLE_GITHUB}" = "true" ] ; then
-  eval "$("${SCRIPT_DIR}"/../../scripts/upload-secrets/manage-github-secrets.sh retrieve)"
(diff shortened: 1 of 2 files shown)


AP-Hunt (approved):
Happy with this. I didn't know we used this script in this way.

PR #536, opened 2022-11-03
Split init-bucket pipeline into update-pipeline and terraform bucket

corlettb:
What
----

Currently the init-bucket job both updates the pipeline and runs the bucket terraform. This means that if you update the terraform pipeline it won't take effect until you rerun the pipeline. This affected our recent terraform upgrade where a 'double tap' of the build pipeline was required.

This change splits out the bucket terraform from the update pipeline to reduce the cases where this 'double tap' would be required.

How to review
-------------

Check you are happy with the pipeline changes.

New pipeline can be viewed [here](https://deployer.dev04.dev.cloudpipeline.digital/teams/main/pipelines/create-bosh-concourse)

Simulated failure that didn't require a 'double tap' can be viewed [here](https://deployer.dev04.dev.cloudpipeline.digital/teams/main/pipelines/create-bosh-concourse/jobs/init-bucket/builds/4#L635ab972:1).

Who can review
--------------

One of the old timers.
---

### DIFF
--- a/concourse/pipelines/create-bosh-concourse.yml
+++ b/concourse/pipelines/create-bosh-concourse.yml
@@ -48,6 +48,7 @@ meta:
 groups:
   - name: all
     jobs:
+      - update-pipeline
       - init-bucket
       - check-for-secrets
       - vpc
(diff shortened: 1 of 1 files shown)


AP-Hunt (approved):

risicle:
So much better.

PR #549, opened 2023-01-16
An empty commit so I can sign it with my own key

AP-Hunt:
What
----

Malcolm's key signs the most recent commit on `main`, but some pipelines haven't run since before his key was added, thus they can't verify the commit (because they can't update to see his key).

How to review
-------------

Did it make no changes?

---


malcgds:
Looks good

PR #554, opened 2023-01-20
[#184266539] Fix create-bosh-concourse in the build environment

corlettb:
What
----

create-bosh-concourse has been "unreliable" on ci build for a long time now.

It turns out that this is because the pipeline doesn't support multiple concourse workers. This is why is worked from concourse-lite with a single worker.

One part of the pipeline gets the workers current ip, then next job adds the ip to the security group and a further job requires this firewall rule to be open. If these jobs happen to run on different concourse worker nodes the pipeline will fail.

This change ensures all workers are granted access to the required endpoints. 

How to review
-------------

Look at the change

Who can review
--------------

Any team member.

---

### DIFF
--- a/terraform/bosh/security_groups.tf
+++ b/terraform/bosh/security_groups.tf
@@ -1,3 +1,15 @@
+data "aws_instances" "concourse_workers" {
+  filter {
+    name   = "tag:instance_group"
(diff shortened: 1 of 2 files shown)


jackjoy-gds (approved):
OK

PR #562, opened 2023-02-16
create-bosh-concourse pipeline: start `pipeline-trigger` after `init-bucket`

risicle:
What
----

In #536 we split the state bucket creation out of the `update-pipeline` job, but kept the starting of the `pipeline-trigger` on the `update-pipeline` job, but this uses the state-bucket to operate and won't work on a clean deployment where the state bucket doesn't yet exist. So move the start of the `pipeline-trigger` to the end of the `init-bucket` job.

How to review
-------------

Deploy to a dev env? If you're really keen, use it to initialize a new dev env but that will probably take half a day...

---

### DIFF
--- a/concourse/pipelines/create-bosh-concourse.yml
+++ b/concourse/pipelines/create-bosh-concourse.yml
@@ -495,15 +495,9 @@ jobs:
                 echo 'self-update-pipeline failed'
                 exit 1
 
-      - put: pipeline-trigger
-        params: {bump: patch}
-
   - name: init-bucket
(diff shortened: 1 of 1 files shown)


corlettb (approved):
My bad. Approved.

risicle:
Ben has since noticed that this fix itself isn't perfect because `init-bucket` doesn't get automatically triggered by an `update-pipeline` run if there's no new `paas-bootstrap` revision involved. Better solutions welcome.

PR #577, opened 2023-03-20
184281154 concourse rebuilds corrupt state

malcgds:
What
----

The [pivotal storey](https://www.pivotaltracker.com/n/projects/1275640/stories/184281154) provides the detail.

The create-bosh-concourse pipeline has been changed such that Terraform is now configured with its own backend and is no longer reliant on the pipeline providing and saving its state via resources.

I have also used the aws cli to pull down a file from an s3 bucket to obtain the current bosh manifest state.

How to review
-------------

1 - Review the code.
2 - Run the branch through a dev env (I've tested this branch on dev01).
3 - After the run, check the size of the files (bosh.tfstate, bosh-manifest-state-(region).json, concourse.tfstate, bootstrap-cyber.tfstate, vpc.tfstate) in the appropriate bucket for the env e.g. for dev01 the bucket would be gds-paas-dev01-state

Who can review
--------------

Not me.

---

### DIFF
--- a/concourse/pipelines/create-bosh-concourse.yml
+++ b/concourse/pipelines/create-bosh-concourse.yml
@@ -209,28 +209,6 @@ resources:
       initial_version: "-"
       initial_content_text: ""
 
-  - name: bootstrap-cyber-tfstate
-    type: s3-iam
-    source:
-      bucket: ((state_bucket))
-      versioned_file: bootstrap-cyber.tfstate
-      region_name: ((aws_region))
-      initial_version: "-"
-      initial_content_text: |
-        {
-            "version": 1,
-            "serial": 0,
-            "modules": [
-                {
-                    "path": [
-                        "root"
-                    ],
-                    "outputs": {},
-                    "resources": {}
-                }
-            ]
-        }
-
   - name: ssh-private-key
     type: s3-iam
     source:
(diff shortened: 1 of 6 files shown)


concourse/pipelines/create-bosh-concourse.yml
```diff
@@ -1125,6 +1113,27 @@ jobs:
         - get: bosh-init-state
         - get: ssh-private-key
 
+        # task to work around potential bosh state corruption on build re-runs - 184281154
```
cadmiumcat, line 1080: I'd suggest making this comment future proof - Imagine an engineer comes along in a year and for some reason we don't have pivotal anymore, what would be a helpful comment?

malcgds: I don't see that as a problem. I believe that the comment before the Pivotal Tracker reference conveys sufficient information so as to mitigate the possibility of Pivotal Tracker becoming unavailable. I could get rid of the story number if that is likely to be confusing but it also appears in the PR name which would likely lead to a similar problem should Pivotal Tracker become unavailable. What do you think?

 If you disagree with my reasoning of course I'll change it.

cadmiumcat: you're right 👍

corlettb (approved):

PR #524, opened 2022-10-11
destroy-bosh-concourse: fix to work with terraform 1.2

risicle:
What
----

This pipeline hasn't worked since #476, where terraform was upgraded from 0.13 -> 1.0. Terraform now has to be run from the root module's directory and the `terraform destroy -force` argument had been removed.

How to review
-------------

Do you really want to run a full env teardown?

---

🚨⚠️ Please do not merge this pull request via the GitHub UI ⚠️🚨

PR #476, opened 2021-10-07
update docker containers to latest hash

LeePorte:
What
----

Update docker containers to get the latest terraform

How to review
-------------

Check the hash matches the latest from https://concourse.build.ci.cloudpipeline.digital/teams/main/pipelines/build-docker-containers/jobs/build-terraform/builds/7

Who can review
--------------

Anyone

---

🚨⚠️ Please do not merge this pull request via the GitHub UI ⚠️🚨

PR #587, opened 2023-03-30
Vagrant fixes after testing

corlettb:
What
----

- Revert [Revert "[#184602228] Fix create-bosh-concourse initial pipeline trigger](https://github.com/alphagov/paas-bootstrap/pull/585/files). After testing this did not cause any problems.
- Fix path issues with update-terraform-providers.sh as part of destroy-bosh-concourse. It looks like this has been broken for some time.
- Ensure the vagrant deploy.sh is idempotent.
- Used fixed versions of the concourse docker image rather than latest.

How to review
-------------

This has been tested against a new environment.

Who can review
--------------

Team elder.

---

### DIFF
--- a/vagrant/deploy.sh
+++ b/vagrant/deploy.sh
@@ -23,16 +23,19 @@ export CONCOURSE_WEB_USER CONCOURSE_WEB_PASSWORD
 export VAGRANT_DEFAULT_PROVIDER="aws"
 export VAGRANT_BOX_NAME="aws_vagrant_box"
 
-if aws ec2 describe-key-pairs --key-name "${VAGRANT_SSH_KEY_NAME}" >/dev/null 2>&1 ; then
-  # Remove the aws key pair
-  aws ec2 delete-key-pair --key-name "${VAGRANT_SSH_KEY_NAME}"
-fi
- 
-# Create the key pair online.
-aws ec2 create-key-pair --key-name "${VAGRANT_SSH_KEY_NAME}" | jq -r ".KeyMaterial" > "${VAGRANT_SSH_KEY}"
+if ! vagrant status 2>&1 | grep "running\ (aws)" >/dev/null; then
+  echo "Concourse Bootstrap VM is not running."
+  if aws ec2 describe-key-pairs --key-name "${VAGRANT_SSH_KEY_NAME}" >/dev/null 2>&1 ; then
+    echo "Cleaning up old key pair..."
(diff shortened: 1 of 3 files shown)


concourse/pipelines/destroy-bosh-concourse.yml
```diff
@@ -576,7 +576,7 @@ jobs:
                 cd paas-bootstrap/terraform/bosh || exit
                 terraform init
 
-                sh paas-bootstrap/terraform/./update-terraform-providers.sh updated-bosh-tfstate/bosh.tfstate
+                sh ../../../paas-bootstrap/terraform/./update-terraform-providers.sh ../../../updated-bosh-tfstate/bosh.tfstate
```
jackjoy-gds: why has the path changed here? was it not working before?

corlettb: This was broken as part of this change:

https://github.com/alphagov/paas-bootstrap/pull/476/files#diff-404545febfe9ecaf3bbdbd502f66358a6fa60b610ea61c8663f7d3d6847e4733R623

There was a partial fix done:

https://github.com/alphagov/paas-bootstrap/pull/524/files#diff-404545febfe9ecaf3bbdbd502f66358a6fa60b610ea61c8663f7d3d6847e4733R261

However it didn't fix it everywhere.

jackjoy-gds (approved):
all good

vagrant/deploy.sh
```diff
@@ -42,10 +45,10 @@ fi
...
 vagrant up
 
-echo "Setting up SSH tunnel to concourse..."
-
-vagrant ssh -- -L 8080:127.0.0.1:8080 -fN
-
+if ! pgrep -f "ssh.*${VAGRANT_SSH_KEY_NAME}.*-L 8080:127.0.0.1:8080 -fN" >/dev/null; then
+  echo "Setting up SSH tunnel to concourse..."
+  vagrant ssh -- -L 8080:127.0.0.1:8080 -fN
+fi
```
risicle, line 53: Could we at least echo a message saying it's skipping setting up a new tunnel in the other case - will make it much more debuggable if for some reason this check is spuriously exiting zero.

corlettb: Done

risicle (approved):
I'm as happy as i can be for something that adds a new resource type

PR #594, opened 2023-04-21
`create-bosh-concourse` pipeline: address race condition for `generate-concourse-config`

risicle:
What
----
If `concourse-terraform` finishes before `generate-bosh-config` and this triggers `generate-concourse-config`, it will pick up an old copy of `bosh-secrets`, not yet updated by `generate-bosh-concourse`. This is bad and can cause severe problems when secrets are being rotated.

This ensures the `pipeline-trigger` doesn't reach `generate-concourse-config` until it has also passed `generate-bosh-config`.

How to review
-------------

This probably needs to be run through an env and the env have its credentials rotated via `drop-bosh-credentials-for-rotation` a la https://team-manual.cloud.service.gov.uk/team/rotating_credentials/#rotating-bosh-credentials-and-certificates. To ensure we test the case where `concourse-terraform` finishes before `generate-bosh-config`, it's probably worth briefly pausing `generate-bosh-config` to delay its start as the release passes it.

We should also test that we can bootstrap an env from concourse-lite with this change, as laborious as it is.

---

### DIFF
--- a/concourse/pipelines/create-bosh-concourse.yml
+++ b/concourse/pipelines/create-bosh-concourse.yml
@@ -1344,7 +1344,7 @@ jobs:
           passed: ['concourse-terraform']
         - get: pipeline-trigger
           trigger: true
-          passed: ['concourse-terraform']
+          passed: ['concourse-terraform', 'generate-bosh-config']
         - get: vpc-tfstate
           passed: ['concourse-terraform']
         - get: concourse-tfstate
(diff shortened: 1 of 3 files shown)


fearoffish (approved):
Simple fix for the race condition. Like it.

risicle:
I have now tried:
 - bringing up a dev env from scratch with this branch (successful)
 - tricking it into running in the problematic order while rotating secrets (it wouldn't let me)

So I think I'm happy with the meat of this PR, but I'd now like someone to put some :eye: on the S3 bucket changes.

jackjoy-gds (approved):
Looks good

PR #606, opened 2023-06-05
Improvements to bosh-cli.sh

whi-tw:
## What

We were stomping on the user's shell, which plays havoc with shell history / scrollback when using zsh.
Instead, we should just create a subshell from `$SHELL`.

Also, use a temporary directory for bosh config - this way, we avoid blatting `~/.bosh/config`, in case someone has something in there that they like.

## How to review

Code review

## Who can review

Not me

---

### DIFF
--- a/scripts/bosh-cli.sh
+++ b/scripts/bosh-cli.sh
@@ -1,6 +1,17 @@
 #!/bin/bash
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


jackjoy-gds (approved):
LGTM

risicle:
This assumes a macos-style `mktemp` and breaks if you're using a linux/gnu `mktemp`.

PR #607, opened 2023-06-07
terraform: add VPC flow logs logging parquet files to a bucket

risicle:
What
----

https://www.pivotaltracker.com/story/show/185202227

This uses a bucket expiry time of 30d as that's the minimum we will be charged for.

The generated parquet files are quite convenient to work with - I've had good success querying them with `duckdb`.

My thinking is to leave this enabled on all envs as the volume it's likely to create on dev will pale in comparison to prod.

This depends on https://github.com/alphagov/paas-aws-account-wide-terraform/pull/360 having been run in the target aws account before deploying.

How to review
-------------

Deploy to a dev env. Go look at the `gds-paas-${var.env}-vpc-flow-logs` bucket it has created filling with parquet files.

I've also tested bringing a dev env all the way down and up again with this successfully.

---

### DIFF
--- a/terraform/vpc/flow-log.tf
+++ b/terraform/vpc/flow-log.tf
@@ -1,0 +1,52 @@
+resource "aws_s3_bucket" "vpc-flow-logs" {
+  bucket = "gds-paas-${var.env}-vpc-flow-logs"
+
+  tags = {
+    Environment = var.env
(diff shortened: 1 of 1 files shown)


fearoffish (approved):
lgtm

PR #643, opened 2023-08-24
load balancers: write logs to global bucket

risicle:
What
----

https://www.pivotaltracker.com/story/show/185739015

This is @whi-tw 's work, barely altered.

This simply sets up the load balancers we use for auxiliary services to log to the account-wide logging bucket added in https://github.com/alphagov/paas-aws-account-wide-terraform/pull/375. As such it depends on that PR.

How to review
-------------

:eyes: Look at dev03? Deploy to your own dev env?

---

🚨⚠️ Please do not merge this pull request via the GitHub UI ⚠️🚨

PR #646, opened 2023-08-29
Revert "load balancers: write logs to global bucket"

risicle:
Reverts alphagov/paas-bootstrap#643

Let's roll this back to unblock pipelines while we figure out the cross-region nonsense.


nimalank7 (approved):
LGTM!

PR #660, opened 2023-09-25
bosh db: upgrade postgresql 11 -> 13

risicle:
What
----
https://www.pivotaltracker.com/story/show/185989513

This requires some trickery and will require some cleanup that can't happen while we still have envs that haven't done the upgrade.

It's also possible that the new `engine_version` will be rejected in different environments which happen to be on different minor versions of postgres 11, where 13.10 isn't a valid upgrade target. 13.10 is a valid upgrade target for 11.19.

Why 13 instead of 14 or 15? We don't really have any 14 or 15 databases at the moment, and it seems better to keep the fleet in line.

How to review
-------------

Deploy to a dev env, see the upgrade happening & bosh continuing to work afterwards.

---

### DIFF
--- a/terraform/bosh/rds.tf
+++ b/terraform/bosh/rds.tf
@@ -28,10 +28,17 @@ resource "aws_security_group" "bosh_rds" {
   }
 }
 
+# TODO remove once all envs have been upgraded past 11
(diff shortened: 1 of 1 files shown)


fearoffish (approved):
deployed to dev05, no issues

